PR #20, opened 2018-04-18
PRO-2423: Add Redis initialization for CNP

tomxelliott:
- Set up in one of two ways based on CNP-Enabled variable

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/app/components/utils.js
+++ b/app/components/utils.js
@@ -44,15 +44,19 @@ exports.getStore = function (useRedis, session) {
     if (useRedis) {
         const ioRedis = require('ioredis');
         const RedisStore = require('connect-redis')(session);
-        const client = ioRedis.createClient(config.redis.port, config.redis.host);
+        const client = ioRedis.createClient(config.redis.port,
+                {'password': config.redis.password,
+                    'host': config.redis.host,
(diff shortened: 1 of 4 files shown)


app/components/utils.js
```diff
@@ -44,16 +44,20 @@ exports.getStore = function (useRedis, session) {
     if (useRedis) {
         const ioRedis = require('ioredis');
         const RedisStore = require('connect-redis')(session);
-        const client = ioRedis.createClient(config.redis.port, config.redis.host);
+        const client = ioRedis.createClient(config.redis.port,
```
reddy1praveen: Hi we need old code as well
use if password is there.

app/healthcheck.js
```diff
@@ -29,7 +29,7 @@ function createPromisesList(services) {
 
 function getServiceHealthUrl(url) {
```
shepda: Perhaps we could remove the duplication from this method by using a ternary for the port, something like this?
```
function getServiceHealthUrl(url) {
    const port = url.port ? `:${url.port}` : '';
    return `${url.protocol}//${url.hostname}${port}'/health'`;
}
```

PR #30, opened 2018-04-26
Pro 869 consolidate config within config js

davejones74:
https://tools.hmcts.net/jira/browse/PRO-869

### Change description ###
Some config is in server.js (app.js), it should be consolidated within config.js
Specifically this has involved moving process.env.* properties into one file config.js.

```
[ ] Yes
[x ] No
```


app/components/api-utils.js
```diff
@@ -4,6 +4,7 @@ const {endsWith} = require('lodash');
 
 const fetch = require('node-fetch');
 const HttpsProxyAgent = require('https-proxy-agent');
+const config = require('app/config.js');
```
shepda: The require statement doesn't need the .js at the end of the file path so this could be changed to 

`const config = require('app/config');`

app.js
```diff
@@ -12,14 +12,14 @@ const routes = require(__dirname + '/app/routes');
 const favicon = require('serve-favicon');
 const bodyParser = require('body-parser');
 const cookieParser = require('cookie-parser');
-const config = require(__dirname + '/app/config.js');
-const utils = require(__dirname + '/app/components/utils.js');
+const config = require(__dirname + '/app/config');
+const utils = require(__dirname + '/app/components/utils');
 const packageJson = require(__dirname + '/package.json');
```
shepda: Can we remove the `.json` extension from this `require` too?

PR #47, opened 2018-05-08
PRO-2536: Add revised Probate specific footer links and their pages

adamb-moj:

https://tools.hmcts.net/jira/browse/PRO-2536

### Change description ###
Add Probate specific footer links and its content.

```
[ ] Yes
[x] No
```


app/resources/en/translation/cookies.json
```diff
@@ -0,0 +1,72 @@
...
+  "title": "Cookies",
+  "heading1": "Cookies",
+  "paragraph1": "A cookie is a small piece of data that&rsquo;s stored on your computer, tablet, or phone when you visit a website. Most websites need cookies to work properly.",
+
+  "howAreCookiesUsedHeading1": "How cookies are used in this service",
+  "howAreCookiesUsedParagraph1": "This service uses them to:",
+  "howAreCookiesUsed-item1": "measure how you use the service so it can be improved",
+  "howAreCookiesUsed-item2": "remember the notifications you&rsquo;ve seen so that you&rsquo;re not shown them again",
+  "howAreCookiesUsed-item3": "temporarily store the answers you give",
+  "howAreCookiesUsedParagraph2": "Find out more about <a rel=\"external noopener\" href=\"http://www.aboutcookies.org/\">how to manage cookies</a>.",
```
shepda: I'd prefer these to be split into sections instead of repeating the section name in the keys, for example:

```
"howAreCookiesUsed": {
  "paragraph1": "...",
  "item1": "...",
  "item2": "...",
  "item3": "...",
  "paragraph2": "..."
}
```

shepda:
The `safe` filter is used when there are html codes in the text that should be displayed as they're entered and not escaped. If we have simple text then we don't need to use `safe`. Please check that `safe` is only being used where it needs to be used.

app/resources/en/translation/cookies.json
adamb-moj: As discussed, this is a possible job for later.  The layout works for these particular pages (and in fact improves it as you say) but the layout will then be different to existing files, some of which would become less easy to read if we followed the same format.

shepda: Yep ok, as it's the same format as the other pages, let's leave it for now.

PR #50, opened 2018-05-11
PRO-2663: Fix to Address Lookup in IE8 to stop Undefined text.

annemariegarcia:

https://tools.hmcts.net/jira/browse/PRO-2663

### Change description ###
Added a polyfill to fix formaction in HTML5

```
[ ] Yes
[x ] No
```

### DIFF
--- a/app/views/includes/scripts.html
+++ b/app/views/includes/scripts.html
@@ -11,3 +11,7 @@
 {% if enableTracking === 'true' %}
 <script src="/public/javascripts/gatracking.js"></script>
 {% endif %}
+
+<!--[if IE 8]>
+<script src="/public/javascripts/ie8polyfill.js"></script>
+<![endif]-->
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,7 +1,5 @@
 node_modules/*
 govuk/*
 public/*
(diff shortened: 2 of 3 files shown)


annemariegarcia:
Added a polyfill to fix formaction in HTML5

app/views/includes/scripts.html
```diff
@@ -10,4 +10,8 @@
 
 {% if enableTracking === 'true' %}
 <script src="/public/javascripts/gatracking.js"></script>
-{% endif %}
\ No newline at end of file
+{% endif %}
+
+<!--[if IE 8]>
```
shepda: We've got this as a bug for IE8 but `formaction` is also not supported by IE9.

Have you tested this with IE9? If not, can you do that to see if we need to enable the polyfill for that browser too?

app/views/includes/scripts.html
```diff
@@ -10,4 +10,8 @@
...
+<script src="/public/javascripts/ie8polyfill.js"></script>
```
shepda: Perhaps a better name for this file would be something like `form-polyfill.js` as it's then linked to the functionality and not then tied to a specific browser?

PR #55, opened 2018-05-15
Pro 2694 multiple payments being taken

davejones74:
https://tools.hmcts.net/jira/browse/PRO-2694

### Change description ###
The payment-status introduced a new value for paymentPending which can now be 'unknown' if the authorisation (s2s) fails to complete. This issue means the application doesn't know if the user has paid or not. 

The solution was to ignore the handlePost for the payment-breakdown page if the paymentPending was 'unknown' - instead the user is redirected back to the payment-status where the check is performed again therefore preventing the user from paying multiple times.  

```
[ ] Yes
[x ] No
```


app/steps/ui/payment/breakdown/index.js
```diff
@@ -54,47 +56,55 @@ module.exports = class PaymentBreakdown extends Step {
...
+                formdata.paymentPending = 'false';
+                delete this.nextStepUrl;
             }
-        } else {
-            formdata.paymentPending = 'false';
-            delete this.nextStepUrl;
 
+        } else {
+            logger.warn('Skipping create payment as authorisation is unknown.');
+            this.nextStepUrl = () => this.steps.PaymentStatus.constructor.getUrl();
```
nigeldunne: We only need the logger statement in this block

davejones74: Ok, you're right. I've removed the nextStepUrl and return as these were unccessary.

PR #69, opened 2018-05-21
PRO-2750 - Added deceased's name and updated test

annemariegarcia:
https://tools.hmcts.net/jira/browse/PRO-2750

### Change description ###
Added Deceased's name using session variable and updated the test for that page.

```
[ ] Yes
[x] No
```

### DIFF
--- a/app/steps/ui/deceased/alias/template.html
+++ b/app/steps/ui/deceased/alias/template.html
@@ -4,8 +4,8 @@
 {% from "widgets/radiobuttons.html" import optionsRadio %}
 
 {% block question %}
-    <h1 class="form-title heading-large">{{content.question}}</h1>
-    <p>{{content.paragraph1}}</p>
+    <h1 class="form-title heading-large">{{content.question | replace('{deceasedName}', fields.deceasedFullName.value) | safe }}</h1>
+    <p>{{content.paragraph1 }}</p>
 {% endblock %}
 
 {% block form_content %}
--- a/app/steps/ui/deceased/alias/template.html
+++ b/app/steps/ui/deceased/alias/template.html
@@ -4,7 +4,7 @@
 {% from "widgets/radiobuttons.html" import optionsRadio %}
 
 {% block question %}
-    <h1 class="form-title heading-large">{{content.question | replace('{deceasedName}', fields.deceasedFullName.value) | safe }}</h1>
+    <h1 class="form-title heading-large">{{content.question | replace('{deceasedName}', fields.deceasedName.value) | safe }}</h1>
     <p>{{content.paragraph1 }}</p>
 {% endblock %}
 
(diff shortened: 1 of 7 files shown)


test/component/deceased/testAlias.js
```diff
@@ -20,7 +23,18 @@ describe('deceased-alias', () => {
     describe('Verify Content, Errors and Redirection', () => {
 
         it('test right content loaded on the page', (done) => {
-            testWrapper.testContent(done, []);
+            set(sessionData, 'deceased.firstName', 'John');
+            set(sessionData, 'deceased.lastName', 'Doe');
```
shepda: We can set the `sessionData` object directly here instead of using `set`, for example:

```
const sessionData = {
  deceased: {
    firstName: 'John',
    surname: 'Doe'
  }
}
```

As we're setting the entire object we don't need any of the convenience `set` gives us and therefore we don't need to require an additional library.

test/component/deceased/testAlias.js
```diff
@@ -1,16 +1,19 @@
 const TestWrapper = require('test/util/TestWrapper'),
+    {set} = require('lodash'),
     DeceasedOtherNames = require('app/steps/ui/deceased/otherNames/index'),
     DeceasedMarried = require('app/steps/ui/deceased/married/index'),
     DeceasedDod = require('app/steps/ui/deceased/dod/index');
 
+
 describe('deceased-alias', () => {
-    let testWrapper;
+    let testWrapper, sessionData;
```
shepda: We don't need to declare `sessionData` here as it can be scoped to the test that uses it.

test/component/deceased/testAlias.js
```diff
@@ -1,16 +1,19 @@
...
     const expectedNextUrlForDeceasedOtherNames = DeceasedOtherNames.getUrl();
     const expectedNextUrlForDeceasedMarried = DeceasedMarried.getUrl();
     const expectedNextUrlForDeceasedDod = DeceasedDod.getUrl();
 
     beforeEach(() => {
         testWrapper = new TestWrapper('DeceasedAlias');
+        sessionData = {};
```
shepda: We don't need to set `sessionData` here as it can be set in the test that uses it.

app/steps/ui/deceased/alias/template.html
```diff
@@ -4,8 +4,8 @@
 {% from "widgets/radiobuttons.html" import optionsRadio %}
 
 {% block question %}
-    <h1 class="form-title heading-large">{{content.question}}</h1>
-    <p>{{content.paragraph1}}</p>
+    <h1 class="form-title heading-large">{{content.question | replace('{deceasedName}', fields.deceasedFullName.value) | safe }}</h1>
```
shepda: As `deceasedName` and `deceasedFullName` refer to the same data it would be nice to have the variables names the same, can we change one of these so they are both the same?

PR #82, opened 2018-05-29
Pro 2704 b cookie banner content missing

adamb-moj:
https://tools.hmcts.net/jira/browse/PRO-2704

### Change description ###
Correct cookie banner so it shows correctly plus other code tidying
    
    - Removed service name from app.js. It now uses the serviceName
    in common.json (this is for multi-language).
    
    - Removed version info from app.js as it wasn't used.
    
    - Removed service object from config.js as it's no longer being used.
    
    - Cookies link updated in test/config.js
    
    - Healthcheck and testHealthcheck now refer to serviceName from
    common.json (note this is the English version of content, as the service
    name written in Welsh in Healthcheck status could confuse).

```
[ ] Yes
[x] No
```

### DIFF
--- a/app/resources/en/translation/common.json
+++ b/app/resources/en/translation/common.json
@@ -12,7 +12,7 @@
   "acceptAndContinue": "Accept and continue",
   "send": "Send",
   "ogl": "All content is available under the <a href=\"https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/\" rel=\"license\">Open Government Licence v3.0</a>, except where otherwise stated",
-  "cookieText": "GOV.UK uses cookies to make the site simpler. <a href=\"#\" title=\"Find out more about cookies\">Find out more about cookies</a>",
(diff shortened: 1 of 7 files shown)


app/resources/en/translation/common.json
```diff
@@ -12,7 +12,7 @@
   "acceptAndContinue": "Accept and continue",
   "send": "Send",
   "ogl": "All content is available under the <a href=\"https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/\" rel=\"license\">Open Government Licence v3.0</a>, except where otherwise stated",
-  "cookieText": "GOV.UK uses cookies to make the site simpler. <a href=\"#\" title=\"Find out more about cookies\">Find out more about cookies</a>",
+  "cookieText": "GOV.UK uses cookies to make the site simpler. <a href=\"/cookies\" rel=\"noopener\" title=\"Find out more about cookies\">Find out more about cookies</a>",
```
shepda: Ideally at some stage it would be good to reference the `/cookies` url from the config as we have with other urls.

Also check if we need to use `noopener` on internal urls.

PR #83, opened 2018-05-29
PRO-2869: Updated package.json to use modules without security issues.

davejones74:
https://tools.hmcts.net/jira/browse/PRO-2869

### Change description ###
Executing yarn test:nsp was complaining at vulnerable versions of tunnel-agent and debug. tunnel-agant was a pain to change as the latest versions all use request 2.79.0 which is tied to the older version of tunnel-agent.

https://github.com/sass/node-sass/issues/2355

However changing node-sass the original package to 4.6.1 allows transitive dependancies to be upgraded so I was then able to install upgrades for debug and tunnel-agent and request modules. 

Changes to package.json are as follows.

(new)    "debug": "^3.1.0",
(removed)    "node-sass": "^4.5.3",
(new)    "node-sass": "^4.6.1",
(removed)    "request": "^2.81.0",
(new)    "request": "^2.87.0",
(new)    "tunnel-agent": "^0.6.0",

```
[ ] Yes
[x] No
```


.gitignore
```diff
@@ -10,10 +10,12 @@
 /output/
 .sonar/
 .sonarlint/
+.project
 /npm-debug.log*
 jsconfig.json
 *.tmp
 coverage
+yarn.lock
```
reddy1praveen: don't add it here

PR #112, opened 2018-06-07
PRO-2839: Add service name to page titles and adapt tests accordingly

adamb-moj:
https://tools.hmcts.net/jira/browse/PRO-2839

### Change description ###
- Page titles moved to layout.html instead of two_thirds_form.html
- Page titles removed from template.html for each page.
- Page title is overridden in a couple of places (e.g. summary/template.html)
- Pa11y tests updated

**NB:  Question marks have been removed from certain title elements in the content files because they 
break the Pa11y tests.  It appears it's a bug with Pa11y.**

```
[ ] Yes
[x] No
```


tomxelliott:
The code looks fine mate, might need to just investigate why the builds aren't working though. :)

test/util/a11y.js
```diff
@@ -8,6 +8,8 @@ const pa11yRun = pa11y({
 });
 
 module.exports = (testPage, title) => {
+    console.log(`Test testPage being tested = ${testPage}`);
+    console.log(`Test title being tested = ${title}`);
```
shepda: Pa11y already provides some output when running the tests, do we need these here as well?

shepda:
Yep good point...I've added a do not merge label

PR #114, opened 2018-06-07
PRO-2590 Add missing list item

annemariegarcia:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/app/steps/ui/startpage/template.html
+++ b/app/steps/ui/startpage/template.html
@@ -17,7 +17,7 @@
     <p>{{content.paragraph2 | safe}}</p>
 
     <div class="panel panel-border-wide">
-        {{content.otherServices | safe}}
+        <p>{{content.otherServices | safe}}</p>
     </div>
 
     <p>{{content.paragraph3 | safe}}</p>
--- a/app/steps/ui/startpage/template.html
+++ b/app/steps/ui/startpage/template.html
@@ -17,7 +17,7 @@
     <p>{{content.paragraph2 | safe}}</p>
 
     <ul class="panel panel-border-wide">
-        {{content.otherServices | safe}}
+        <li>{{content.otherServices | safe}}</li>
     </ul>
 
     <p>{{content.paragraph3 | safe}}</p>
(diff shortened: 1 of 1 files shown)


app/steps/ui/startpage/template.html
```diff
@@ -17,7 +17,7 @@ <h1 class="form-title heading-large">{{content.header}}</h1>
     <p>{{content.paragraph2 | safe}}</p>
 
     <ul class="panel panel-border-wide">
-        {{content.otherServices | safe}}
+        <li>{{content.otherServices | safe}}</li>
     </ul>
```
shepda: As there is only one list item here, do we need to use a list? Perhaps we'd be better to use a generic `<div>` tag as is done in the GOV.UK elements. See http://govuk-elements.herokuapp.com/typography/ under 'Inset text' for an example.

shepda:
Your commit message doesn't conform to the commit message format we're now using, it should be something like 'PRO-2590: Add missing list item'. 

Can you please check you have the git hooks installed, which will run specific checks before commits are accepted. 

You can install these by running the following commands from the root of the app:

```
ln -s ../../pre-commit.sh .git/hooks/pre-commit
ln -s ../../commit-msg.sh .git/hooks/commit-msg
```

See the `README.md` for more information.

app/steps/ui/startpage/template.html
```diff
@@ -16,9 +16,9 @@ <h1 class="form-title heading-large">{{content.header}}</h1>
     <p>{{content.paragraph1 | safe}}</p>
     <p>{{content.paragraph2 | safe}}</p>
 
-    <ul class="panel panel-border-wide">
+    <div class="panel panel-border-wide">
         {{content.otherServices | safe}}
-    </ul>
+    </div>
```
shepda: We should add a `<p>` tag around the content within the `<div>` to denote that this is a paragraph of text within the DOM.

This will also keep the markup consistent with the GOV.UK elements.

app/steps/ui/startpage/template.html
adamb-moj: Yes, we do in this instance.  This is due to the content containing two hyperlinks.  If it only contained one then it could be a div or such.  I've changed it back so it now passes Pa11y.

PR #115, opened 2018-06-08
PRO-3045: Adding health status UP to s2s-stub for verifying app healt…

davejones74:
https://tools.hmcts.net/jira/browse/PRO-3045

### Change description ###
Basic change to add a health endpoint to the s2s-stub. This is required to get a health status for all probate applications to be 'UP'. 

The trigger for this is the submit service includes an external JAR file which calls the s2s service health endpoint - not sure if this is necessary - however in development this endpoint in the s2s-stub doesn't exist to the submit service is always down. Adding the health endpoint means the submit service is now healthy.

Alternatively the s2s dependancy in the submit service could also be disabled although I'm not entirely sure why the submit service is using this service if the frontend calls the s2s service directly.

```
[ ] Yes
[x] No
```

### DIFF
--- a/test/service-stubs/idam.js
+++ b/test/service-stubs/idam.js
@@ -37,6 +37,15 @@ router.post('/lease', function (req, res) {
 
 });
 
+router.get('/health', function (req, res) {
+    console.log(req.headers);
+    console.log(req.body);
+    
+	res.status(200);
+    res.setHeader('Content-Type', 'application/json');
+    res.send(JSON.stringify({ status: 'OK' }));
+})
+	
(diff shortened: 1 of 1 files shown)


shepda:
We have a pre-commit hook which should pick up linting errors when a commit is created. 

Can you please check you have the git hooks installed, which you can do by running the following commands from the root of the app:

```
ln -s ../../pre-commit.sh .git/hooks/pre-commit
ln -s ../../commit-msg.sh .git/hooks/commit-msg
```

See the `README.md` for more information.

PR #121, opened 2018-06-11
[PREVIEW] PRO-1941: Adding validation on checkboxes and corresponding error mes…

harjeetkalsi:
DO NOT MERGE 

https://tools.hmcts.net/jira/browse/PRO-1941

### DIFF
--- a/app/resources/en/translation/executors/dealingwithestate.json
+++ b/app/resources/en/translation/executors/dealingwithestate.json
@@ -13,8 +13,8 @@
         "message": "Select at least one executor"
       },
       "invalid":{
-        "summary": "A maximum of 4 executors can named on the grant of probate. ",
-        "message": "You can not select more than 4 executors. "
+        "summary": "A maximum of 4 executors can be named on the grant of probate.",
+        "message": "You can not select more than 4 executors."
     }
 
   }
--- a/test/component/executors/testExecutorsDealing.js
+++ b/test/component/executors/testExecutorsDealing.js
@@ -55,7 +55,7 @@ describe('executors-dealing-with-estate', () => {
                 });
         });
 
-        it('test errors message displayed for more than 4 applicants', (done) => {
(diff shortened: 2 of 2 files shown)


app/resources/en/translation/executors/dealingwithestate.json
```diff
@@ -11,7 +11,12 @@
       "required": {
         "summary": "Select at least one executor",
         "message": "Select at least one executor"
-      }
+      },
+      "invalid":{
+        "summary": "A maximum of 4 executors can named on the grant of probate. ",
```
shepda: I think we need to add 'be' in between 'can' and 'named.

test/component/executors/testExecutorsDealing.js
```diff
@@ -48,11 +48,24 @@ describe('executors-dealing-with-estate', () => {
...
                     const data = {
-                        executorsApplying: 'many clouds'
+                        executorsApplying: ['many clouds']
                     };
                     testWrapper.pageUrl = testWrapper.pageToTest.constructor.getUrl();
                     testWrapper.testRedirect(done, data, expectedNextUrlForExecAlias);
                 });
         });
+
+        it('test errors message displayed for more than 4 applicants', (done) => {
```
shepda: The validation is for the additional applicants (ie: excluding the main applicant) so perhaps this test description should be re-worded to:

`test error message displayed for more than 3 additional applicants`

Otherwise it's slightly misleading because you have 4 executors in the data.

harjeetkalsi: good point!

app/resources/en/translation/executors/dealingwithestate.json
harjeetkalsi: yes, good spot

venumadhavancha:
Tested this on preview env and it works fine.

PR #125, opened 2018-06-13
Pro 1699 save and close

tomxelliott:

- [x] tests have been updated / new tests has been added (if needed)

https://tools.hmcts.net/jira/browse/PRO-1699

### Change description ###
Implement Save and Close.
2/3 form updated to include save and close link provided it is not part of the multiple-execs pages.

```
[ ] Yes
[x] No
```

### DIFF
--- a/app/steps/ui/signout/index.js
+++ b/app/steps/ui/signout/index.js
@@ -20,7 +20,8 @@ module.exports = class SignOut extends Step {
                 throw new Error('Error while attempting to sign out of IDAM.');
             }
             delete req.cookies;
-            req.session.destroy();
+            delete req.sessionID;
+            delete req.session;
             return ctx;
         })
         .catch(err => {
--- a/app/views/includes/two_thirds_form.html
+++ b/app/views/includes/two_thirds_form.html
@@ -20,7 +20,7 @@
                 <input type="hidden" name="_csrf" value="{{csrfToken}}">
                 {% block form_content %}{% endblock %}
             </form>
-            {% block save_and_close %}<a href="/sign-out">Save and close</a>{% endblock %}
+            {% block save_and_close %}<a href="/sign-out">{{common.saveAndClose}}</a>{% endblock %}
             {% block executor_summary %}{% endblock %}
         </div>
     </div>
--- a/test/component/testSignOut.js
+++ b/test/component/testSignOut.js
@@ -1,5 +1,6 @@
 'use strict';
 const TestWrapper = require('test/util/TestWrapper');
+const commonContent = require('app/resources/en/translation/common');
 
 describe('sign-out', () => {
     let testWrapper;
@@ -18,6 +19,13 @@ describe('sign-out', () => {
             const excludeKeys = [];
 
             testWrapper.testContent(done, excludeKeys, {});
+        });
+
+        it('test save and close link is not displayed on the page', (done) => {
+            const playbackData = {};
+            playbackData.saveAndClose = commonContent.saveAndClose;
+
+            testWrapper.testContentNotPresent(done, playbackData);
         });
     });
 });
--- a/test/unit/testSignOut.js
+++ b/test/unit/testSignOut.js
@@ -1,6 +1,5 @@
 'use strict';
 const initSteps = require('app/core/initSteps');
-const expect = require('chai').expect;
 const assert = require('chai').assert;
 const sinon = require('sinon');
 const when = require('when');
@@ -26,15 +25,12 @@ describe('Sign-Out', function () {
                     payloadVersion: '4.1.0',
                     applicantEmail: 'test@email.com'
                 }
-            },
+            }
         };
-
-        req.session.destroy = sinon.spy();
-        req.cookies = sinon.spy();
 
         signOut.getContextData(req).then(() => {
             assert.isUndefined(req.cookies);
-            expect(req.session.destroy).to.have.been.calledOn(req.session);
+            assert.isUndefined(req.session);
             done();
         });
     });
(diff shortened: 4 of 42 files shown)


test/unit/testSignOut.js
```diff
@@ -0,0 +1,40 @@
...
+                    delete this.session;
+                }
+            },
+        };
+
+        co(function* () {
+            yield signOut.getContextData(req);
+            assert.isNull(req.cookies);
+            assert.isNull(req.session);
+        });
```
shepda: If you comment out either of the `delete req.cookies;` or `req.session.destroy();` lines then the tests don't fail.

- I don't think we need to use `co` here but do this with `.then()`. 
- We can use `sinon.spy()` to spy on the req to check if the values have been removed.
- It also looks like we don't need to mock `req.session.destroy()`.

Now when you comment out either of the delete/destroy lines above the tests fail.

For example:

```
const initSteps = require('app/core/initSteps');
const expect = require('chai').expect;
const sinon = require('sinon');
const when = require('when');
const services = require('app/components/services');

describe.only('Sign-Out', function () {
    const steps = initSteps([`${__dirname}/../../app/steps/action/`, `${__dirname}/../../app/steps/ui`]);

    it('test authToken, cookies and session data have been removed from the session', (done) => {
        const signOut = steps.SignOut;
        const signOutStub = sinon.stub(services, 'signOut');
        signOutStub.returns(when(200));

        const req = {
            cookies: {
                _ga: 'dummy_ga',
                _gid: 'dummy_gid',
                _gat: '1'
            },
            session: {
                form: {
                    payloadVersion: '4.1.0',
                    applicantEmail: 'test@email.com'
                }
            },
        };

        req.session.destroy = sinon.spy();
        req.cookies = sinon.spy();

        signOut.getContextData(req).then(() => {
            expect(req.cookies).to.equal(undefined);
            expect(req.session.destroy).to.have.been.calledOn(req.session);
            done();
        });
    });
});
```

app/steps/ui/signout/index.js
```diff
@@ -0,0 +1,29 @@
+const Step = require('app/core/steps/Step');
+const config = require('app/config');
+const logger = require('app/components/logger')('Init');
+const services = require('app/components/services');
+const SECURITY_COOKIE = '__auth-token-' + config.payloadVersion;
```
shepda: Can we use a template literal here instead of concatenating the strings? For example:

```
const SECURITY_COOKIE = `__auth-token-${config.payloadVersion}`;
```

app/resources/en/translation/signout.json
```diff
@@ -0,0 +1,6 @@
+{
+  "title": "Probate - You&rsquo;ve signed out",
```
shepda: Following the work in Adam's story PRO-2839 we're currently adding a suffix of '- Apply for Probate' to the end of each title so we'd be better to remove the 'Probate - ' here so we get a title of:

'You've signed out - Apply for Probate' 

rather than:

'Probate - You've signed out - Apply for Probate'

app/steps/ui/signout/template.html
```diff
@@ -0,0 +1,18 @@
+{% extends "includes/layout.html" %}
+
+{% block page_title %}
+{{ content.title | safe}}
+{% endblock %}
```
shepda: Following the work in Adam's story PRO-2839 the page title is now handled in the layout template so we won't need to use the `page_title` block here and it can be removed.

app/views/includes/two_thirds_form.html
```diff
@@ -20,6 +20,7 @@
                 <input type="hidden" name="_csrf" value="{{csrfToken}}">
                 {% block form_content %}{% endblock %}
             </form>
+            {% block sign_out %}<a href="/sign-out">Save and close</a>{% endblock %}
```
shepda: Can we call this block `save_and-close` rather than `sign_out` as it will keep it consistent with the function of the link?

test/component/testSignOut.js
```diff
@@ -0,0 +1,26 @@
...
+        testWrapper.destroy();
+    });
+
+    describe('Verify Content, Errors and Redirection', () => {
+
+        it('test right content loaded on the page', (done) => {
+            const excludeKeys = [];
+            const contentData = {
+                idamLoginLink: config.services.idam.loginUrl
+            };
```
shepda: Do we need to define this `contentData` as there doesn't appear to be any content that needs to be replaced?

test/unit/testSignOut.js
```diff
@@ -0,0 +1,40 @@
+const initSteps = require('app/core/initSteps');
+const expect = require('chai').expect;
+const assert = require('chai').assert;
+const sinon = require('sinon');
+const when = require('when');
+const services = require('app/components/services');
+
+describe.only('Sign-Out', function () {
```
shepda: Can we remove the `.only` from this test?

test/unit/testSignOut.js
```diff
@@ -0,0 +1,40 @@
...
+
+    const steps = initSteps([`${__dirname}/../../app/steps/action/`, `${__dirname}/../../app/steps/ui`]);
+
+    it.only('test authToken, cookies and session data have been removed from the session', (done) => {
```
shepda: Can we remove the `.only` from this test?

test/unit/testSignOut.js
```diff
@@ -0,0 +1,40 @@
...
+        req.session.destroy = sinon.spy();
+        req.cookies = sinon.spy();
+
+        signOut.getContextData(req).then(() => {
+            assert.isUndefined(req.cookies);
+            expect(req.session.destroy).to.have.been.calledOn(req.session);
+            done();
+        });
+    });
+});
```
shepda: Can we add a unit test for `getUrl()` also?

PR #126, opened 2018-06-18
PRO-3040: Implement sign out link in the nav bar on each relevant pag…

tomxelliott:
…e of the app

	  - since we don't want users to see the sign out link on certain pages an alternate nav bar template has been implemented
	  - the no-sign-out-link template overrides the usual template when we don't want to see a signout link
	  - pages include all co-applicant pages, pin pages and the sign out page
	  - created a new css class to handle the new sign out link to ensure that it is properly placed

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/app/assets/sass/application-common.scss
+++ b/app/assets/sass/application-common.scss
@@ -86,7 +86,7 @@ div.declaration-header {
   float: right;
   color: $white;
   text-decoration: none; }
-  @media (min-width: 641px) {
+  @media (tablet) {
     #global-header .header-proposition #sign-out-header {
       @include bold-19;
       margin-top: 2px; } }
--- a/app/assets/sass/application-common.scss
+++ b/app/assets/sass/application-common.scss
@@ -81,24 +81,19 @@ div.declaration-header {
 }
 
 #global-header .header-proposition #sign-out-header {
-  font-family: "nta", Arial, sans-serif;
-  text-transform: none;
-  font-size: 18px;
-  line-height: 1.2;
+  @include bold-16;
   font-weight: bold;
   float: right;
-  margin-right: 40px;
-  color: white;
+  color: $white;
   text-decoration: none; }
   @media (min-width: 641px) {
     #global-header .header-proposition #sign-out-header {
-      font-size: 18px;
-      line-height: 1.25; } }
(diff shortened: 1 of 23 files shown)


app/assets/sass/application-common.scss
```diff
@@ -74,8 +74,31 @@ li.task-item {
...
+  line-height: 1.2;
+  font-weight: bold;
+  float: right;
+  margin-right: 40px;
+  color: white;
+  text-decoration: none; }
+  @media (min-width: 641px) {
+    #global-header .header-proposition #sign-out-header {
+      font-size: 18px;
+      line-height: 1.25; } }
```
shepda: Can we remove `font-family`, `text-transform`, `line-height` (from both the main rule and the media query) and `margin-right` so they are inherited from the cascade.

For the `font-size`, can we use `@include bold-16` for the main rule and `@include bold-24` for the media query. This will use the GOVUK defined mixins for the size and weight and follows the css for `#proposition-header` to make the `font-size` larger by specifying the mobile size in the main rule and overriding this with a larger size in the media query.

We should also use the `$white` GOVUK sass variable for the colour instead of specifying the value.

For example:

```
#global-header .header-proposition #sign-out-header {
  @include bold-16;
  float: right;
  color: $white;
  text-decoration: none; 
}
@media (min-width: 641px) {
  #global-header .header-proposition #sign-out-header {
    @include bold-24;
    line-height: 1.25; 
  } 
}
```

Can you also check that the base line of `#sign-out-header` is the same as `#proposition-header` for screens sizes smaller and larger than 641px.

app/assets/sass/application-common.scss
```diff
@@ -74,8 +74,27 @@ li.task-item {
...
   margin-bottom: 5px;
+}
+
+#global-header .header-proposition #sign-out-header {
+  @include bold-16;
+  margin-top: 4px;
+  float: right;
+  color: $white;
+  text-decoration: none; }
+  @media (min-width: 641px) {
```
shepda: GDS have a `tablet` variable that will handle the `min-width: 641px` so we'd be better to use that because if they change the width in future we'd get those changes automatically. For example:

`@media(tablet)`

PR #155, opened 2018-06-27
Pro 3031 number task list items

annemariegarcia:
https://tools.hmcts.net/jira/browse/PRO-3031

### Change description ###
Adding list item styling to the task list, so the sequential order is shown. 

```
[ ] Yes
[x ] No
```


app/steps/ui/tasklist/includes/copies.html
```diff
@@ -1,7 +1,7 @@
 {% set previousTaskStatus = fields.previousTaskStatus.value.CopiesTask %}
 
 <li class="task-item">
-    <span class="heading-medium">{{content.copiesTask}}</span>
+    <h3 class="heading-medium">{{content.copiesTask}}</h3>
```
shepda: It would be more semantically correct for these headings to be `h2` as in your previous commit.

app/assets/sass/application-common.scss
```diff
@@ -17,69 +17,65 @@
...
-          margin-top: 0;
-        }
-      }
-      .task-unavailable {
-        background-color: $grey-2;
-        color: $black;
-        float: right;
+      @include media(tablet) {
+        float: left;
+        width: 66.6%;
```
shepda: The GDS breakpoint variables are:

```
$tablet-breakpoint: 641px !default;
$desktop-breakpoint: 769px !default;
```

A) Ideally we should have the title and tag on the same line for the desktop size, at the moment it appears to break before the browser is reduced to this size at around `785px`. It appears that reducing the title width to `65%` would fix this...is `66.6%` from the GDS styling or a width you came up with?

For example, from this:

screen shot 2018-06-29 at 12 09 51

To this:

screen shot 2018-06-29 at 12 10 16


  Can you also check that the title is fully indented so it doesn't appear as in the second example above.

B) When the tag breaks to the next line for tablet screen sizes and smaller I think the following should happen:

- The title becomes full with.
- The top and bottom margin around the tag increases, perhaps by `20px`.
- We should consider aligning the tag to the left.

For example, from this:

screen shot 2018-06-29 at 11 56 08

To this:

screen shot 2018-06-29 at 11 56 49

And this:

screen shot 2018-06-29 at 11 40 07

To this:

screen shot 2018-06-29 at 11 40 55

PR #156, opened 2018-06-27
Pro 3031 tasklist numbering

annemariegarcia:
JIRA link (if applicable)
https://tools.hmcts.net/jira/browse/PRO-3031

Change description
Adding list item styling to the task list, so the sequential order is shown.


shepda:
Closed because we're now using https://github.com/hmcts/probate-frontend/pull/155 instead.

PR #167, opened 2018-07-03
PRO-2755: Fix error summary href on deceased other names screen

adamb-moj:
https://tools.hmcts.net/jira/browse/PRO-2755

### Change description ###
The deceased other names fields originally had square brackets in their
id and name attributes in the HTML.  However this caused the Codecept
methods to fail so the square brackets were removed from the id.

However, the error summary wasn't changed and thus still included the
brackets in the href, and therefore the href linking to the id was lost.

This change fixes that.

NB: The square brackets were left in the name deliberately because the
name attribute is how the data is moved around the HTML.

```
[ ] Yes
[x] No
```


adamb-moj:
NB:  otherNameErros has been fixed in another ticket.

PR #171, opened 2018-07-04
PRO-3165: Fix to handle S2S failures prior to making a payment.

davejones74:
https://tools.hmcts.net/jira/browse/PRO-3165

### Change description ###
I've cleared the paymentPending and creatingPayment flags (set to null) if there is an authorise failure prior to the createPayment call. This change is required because authorise failures cause the application to get stuck and the user ends up in an endless loop. 

In addition to the flags being cleared, I've also introduced a new unit test. The payment stub was updated to handle undefined paymentId's and send back a HTTP 500 error as per the real service. Because of this I had to update the component test data to include a paymentId which before it had ignored and the stub just returned a successful response.

```
[ ] Yes
[x] No
```

### DIFF
--- a/test/unit/testPaymentBreakdown.js
+++ b/test/unit/testPaymentBreakdown.js
@@ -8,14 +8,14 @@ const sinon = require('sinon');
 
 describe('PaymentBreakdown', () => {
     const steps = initSteps([`${__dirname}/../../app/steps/action/`, `${__dirname}/../../app/steps/ui`]);
-    let s2sAuthoriseStub;
+    let authoriseStub;
 
     beforeEach(function () {
-        s2sAuthoriseStub = sinon.stub(services, 'authorise');
+        authoriseStub = sinon.stub(services, 'authorise');
     });
 
     afterEach(function () {
-        s2sAuthoriseStub.restore();
+        authoriseStub.restore();
     });
 
     describe('handleGet', () => {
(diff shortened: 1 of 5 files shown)


test/unit/testPaymentBreakdown.js
```diff
@@ -54,20 +69,44 @@ describe('PaymentBreakdown', () => {
...
             })
-                .catch((err) => {
-                    done(err);
-                });
+            .catch((err) => {
+                done(err);
+            });
+        });
+
+        it('sets paymentPending and createPayment to undefined if authorise fails before createPayment', (done) => {
```
shepda: In this scenario `paymentPending` and `createPayment` are being set to `null` not `undefined`. Can we update the test description to reflect that?

test/unit/testPaymentBreakdown.js
```diff
@@ -2,10 +2,25 @@
...
+        s2sStub = require('test/service-stubs/idam');
+        payStub = require('test/service-stubs/payment');
+    });
+
+    after(() => {
+        s2sStub.close();
+        payStub.close();
+        delete require.cache[require.resolve('test/service-stubs/idam')];
+        delete require.cache[require.resolve('test/service-stubs/payment')];
+    });
```
shepda: We don't need to use the stubs in these unit tests, instead we can mock the service calls to return the data that we expect.

test/unit/testPaymentStatus.js
```diff
@@ -47,9 +53,8 @@ describe('PaymentStatus', function () {
         });
 
         it('Should not create an error if sendApplication succeeds', function (done) {
-
+            s2sAuthoriseStub.returns(Promise.resolve({name: 'Success'}));
```
shepda: This stub should be mocking `services.submitApplication()` rather than `services.authorise()` because the `result` that `services.submitApplication()` returns will ensure that an error is not returned by `sendApplication()` and `formdata` is persisted.

You'll need to mock `services.saveFormData()` so it returns a valid success value and then `sendApplication()` should return the correct `errors` value (none in this case). In this case the app will log an error with `logger.info('Successfully persisted user data');` so you could `spy` on `logger` to ensure that `info()` is called on it with the correct value.

You could also write another test for the case that `services.saveFormData()` and therefore `persistFormData()` returns an error. In this case you'd need to do similar to the test above.

PR #241, opened 2018-07-26
[PREVIEW] PRO-3344: Changing capital letters and spacing for tasklist "Complete…

harjeetkalsi:
https://tools.hmcts.net/jira/browse/PRO-3344

### DIFF
--- a/app/assets/sass/application-common.scss
+++ b/app/assets/sass/application-common.scss
@@ -55,7 +55,7 @@ li.task-item {
       float: left;
       margin-bottom: 10px;
       text-transform: none;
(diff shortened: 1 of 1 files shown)


app/assets/sass/application-common.scss
```diff
@@ -54,6 +54,8 @@ li.task-item {
       color: $black;
       float: left;
       margin-bottom: 10px;
+      text-transform: none;
+      letter-spacing: 0px;
```
shepda: We don't need to specify the units when there is a `0` value 🙂

PR #283, opened 2018-09-11
Pro 3361 multiple applicants update documents page

tomxelliott:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/app/wrappers/Executors.js
+++ b/app/wrappers/Executors.js
@@ -90,6 +90,12 @@ class Executors {
     executorsRemoved() {
         return this.executorsData.executorsRemoved || [];
     }
+
+    executorsAliasList() {
+        return this.executorsList
+            .filter(executor => executor.aliasReason === 'Change by deed poll' || executor.currentNameReason === 'Change by deed poll')
+            .map(executor => executor.alias || executor.currentName);
+    }
 }
 
 module.exports = Executors;
--- a/app/wrappers/Executors.js
+++ b/app/wrappers/Executors.js
@@ -91,7 +91,7 @@ class Executors {
         return this.executorsData.executorsRemoved || [];
     }
 
-    executorsAliasList() {
+    executorsNameChangedByDeedPoll() {
         return this.executorsList
             .filter(executor => executor.aliasReason === 'Change by deed poll' || executor.currentNameReason === 'Change by deed poll')
             .map(executor => executor.alias || executor.currentName);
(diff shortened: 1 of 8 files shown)


app/wrappers/Executors.js
```diff
@@ -90,6 +90,12 @@ class Executors {
     executorsRemoved() {
         return this.executorsData.executorsRemoved || [];
     }
+
+    executorsAliasList() {
```
shepda: This method is only returning executors who have changed their name by deed poll so can we rename it to reflect that?

app/wrappers/Executors.js
```diff
@@ -90,6 +90,12 @@ class Executors {
...
+        return this.executorsList
+            .filter(executor => executor.aliasReason === 'Change by deed poll' || executor.currentNameReason === 'Change by deed poll')
```
shepda: Can we use the same key for both the main applicant and the other applicants rather than having `aliasReason` for one and `changeNameReason` for the other?

app/wrappers/Executors.js
```diff
@@ -90,6 +90,12 @@ class Executors {
...
+            .map(executor => executor.alias || executor.currentName);
```
shepda: Can we use the same key for both the main applicant and the other applicants rather than having `alias` for one and `changeName` for the other?

test/component/testDocumentsPage.js
```diff
@@ -1,7 +1,9 @@
+// eslint-disable-line max-lines
 const TestWrapper = require('test/util/TestWrapper');
 const config = require('app/config');
 const ThankYou = require('app/steps/ui/thankyou/index.js');
 const ihtContent = require('app/resources/en/translation/iht/method');
+const {assert} = require('chai');
```
shepda: `assert` is not being used in the code and is causing a linting error, please remove it.

Do you have the git hooks set up? Details are in the [README file](
https://github.com/hmcts/probate-frontend/blob/master/README.md)

app/wrappers/Executors.js
tomxelliott: As discussed, happy to change but due to the fact both values mean different thing for the main executor and the secondary executors could get confusing if they had the same value. Will leave as is for now, but definitely raise with Femi when he is back.

app/wrappers/Executors.js
tomxelliott: As discussed, happy to change but due to the fact both values mean different thing for the main executor and the secondary executors could get confusing if they had the same value. Will leave as is for now, but definitely raise with Femi when he is back.

app/wrappers/Executors.js
shepda: Sure ok

app/wrappers/Executors.js
shepda: Sure ok

PR #302, opened 2018-09-20
Pro 3080 reorder executors questions

LucaDelBuonoHMCTS:
https://tools.hmcts.net/jira/browse/PRO-3080

### Change description ###
Also implements the feature toggling of the screeners questions in:
https://tools.hmcts.net/jira/browse/PRO-3724

```
[  ] Yes
[X] No
```


app/steps/ui/executors/number/index.js
```diff
@@ -44,13 +45,32 @@ class ExecutorsNumber extends ValidationStep {
...
-    nextStepOptions() {
-        const nextStepOptions = {
+    handlePost(ctx, errors, formdata, session, hostname, featureToggles) {
+        ctx.isToggleEnabled = FeatureToggle.isEnabled(featureToggles, 'screening_questions');
+
+        return [ctx, errors];
+    }
+
+    nextStepOptions(ctx) {
+        if (ctx.isToggleEnabled) {
```
tomxelliott: Is there a way we could assign the value for 'choice' based on the value of ctx.isToggleEnabled, like this:
const choiceValue = ctx.isToggleEnabled ? 'oneExecutorToggleOn' : 'oneExecutor';

tomxelliott: That way we avoid the if block and the code reuse of the options object.

test/unit/testExecutorNumber.js
```diff
@@ -1,11 +1,84 @@
...
+            done();
+        });
+    });
+
+    describe('nextStepOptions()', () => {
+        it('should return the correct options when the FT is off', (done) => {
+            const ctx = {
+                isToggleEnabled: false
+            };
+            const nextStepOptions = ExecutorsNumber.nextStepOptions(ctx);
```
tomxelliott: This test shouldn't really require any refactoring to account for the change suggested above since the logic is the same.

test/unit/testExecutorNumber.js
```diff
@@ -1,11 +1,84 @@
...
+                    choice: 'oneExecutor'
+                }]
+            });
+            done();
+        });
+
+        it('should return the correct options when the FT is on', (done) => {
+            const ctx = {
+                isToggleEnabled: true
+            };
```
tomxelliott: This test shouldn't really require any refactoring to account for the change suggested above since the logic is the same.

app/steps/ui/executors/number/index.js
LucaDelBuonoHMCTS: Yeah I guess we could but this code is only temporary and is getting removed in PRO-3799.

test/unit/testExecutorNumber.js
LucaDelBuonoHMCTS: Again, lots of this stuff will get removed in PRO-3799.

app/steps/ui/executors/number/index.js
shepda: Yep, I think thats fair enough, it's only temporary code and perhaps makes it easier to remove when we clear up the toggle

PR #343, opened 2018-10-03
[PREVIEW] Pro 3358 secondary executors alias reason for name change

tomxelliott:

### Change description ###

```
[ ] Yes
[ ] No
```


test/unit/testExecutorCurrentName.js
```diff
@@ -0,0 +1,150 @@
...
+                options: [
+                    {key: 'continue', value: true, choice: 'continue'},
+                ],
+            });
+            done();
+        });
+
+    });
+
+    describe('action()', () => {
```
shepda: Instead of having 2 tests here for `action()`, we could have 1 test which sets all the params to be removed and checks that they've been removed from the returned `ctx` value.

Params to be checked are:

```
ctx.index;
ctx.currentName;
ctx.continue;
ctx.isToggleEnabled;
```

test/unit/testExecutorCurrentNameReason.js
```diff
@@ -0,0 +1,241 @@
...
+            const param = 1;
+            const ExecutorCurrentNameReason = steps.ExecutorCurrentNameReason;
+            const url = ExecutorCurrentNameReason.constructor.getUrl(param);
+
+            expect(url).to.equal(executorCurrentNameReasonPath + param);
+            done();
+        });
+    });
+
+    describe('getContextData()', () => {
```
shepda: We should also have a unit tests for `getContextData()` to check the following 2 scenarios:

```
} else if (req.params && req.params[0] === '*') {
```
and 
```
} else if (startsWith(req.path, path)) {
```

test/unit/testExecutorCurrentNameReason.js
```diff
@@ -0,0 +1,241 @@
...
+                    isApplying: true
+                }, {
+                    isApplying: true,
+                    currentNameReason: 'marriage'
+                }],
+                index: 1
+            };
+        });
+
+        it('returns the ctx with currentNameReason', (done) => {
```
shepda: This test appears to be a duplicate of the one below.

There is only one positive scenario in this method so we should only need one test to test the returned params.

test/unit/testExecutorCurrentNameReason.js
```diff
@@ -0,0 +1,241 @@
...
+            const ExecutorCurrentNameReason = steps.ExecutorCurrentNameReason;
+            const [ctx] = ExecutorCurrentNameReason.handleGet(testCtx);
+
+            assert.isUndefined(ctx.currentNameReason);
+            assert.isUndefined(ctx.otherReason);
+            done();
+        });
+    });
+
+    describe('handlePost()', () => {
```
shepda: It looks like we need a unit test for `handlePost()` to cover the following scenario

```
if (ctx.currentNameReason !== 'other') {
    delete ctx.list[ctx.index].otherReason;
}
```

test/unit/testExecutorCurrentNameReason.js
```diff
@@ -0,0 +1,241 @@
...
+                options: [
+                    {key: 'continue', value: true, choice: 'continue'},
+                ],
+            });
+            done();
+        });
+    });
+
+    describe('action()', () => {
+        it('removes the correct values from the context', (done) => {
```
shepda: This test should also check that `ctx.index` has been removed from the returned object.

test/unit/testExecutorOtherName.js
```diff
@@ -0,0 +1,46 @@
+/* eslint max-lines: ["error", 500] */
```
shepda: Please add `'use strict;'` as the first line of this file. See https://github.com/hmcts/probate-tech-docs/blob/master/javascript-coding-style.md#strict-mode for an example

test/unit/testExecutorOtherName.js
```diff
@@ -0,0 +1,46 @@
...
+
+const initSteps = require('app/core/initSteps');
+const {expect, assert} = require('chai');
+
+describe('ExecutorsWithOtherNames', () => {
+    const steps = initSteps([__dirname + '/../../app/steps/action/', __dirname + '/../../app/steps/ui']);
+
+    describe('pruneFormData', () => {
+
+        it('test that currentNameReason is deleted when executor has no other name', () => {
```
shepda: These tests can be combined as there is only one conditional in this method we should only need one test to test the returned params.

PR #344, opened 2018-10-03
[PREVIEW] PRO-3794 - Applications are being submitted with incomplete data

shepda:

https://tools.hmcts.net/jira/browse/PRO-3794

### Change description ###

Instead of creating default session data if the session data doesn't exist, we're now checking if the user has saved data before creating a default session.

```
[ ] Yes
[x] No
```


shepda:
no longer needed

PR #347, opened 2018-10-04
[PREVIEW] PRO-3858: Update to add regId as applicantEmail in formdata when crea…

davejones74:
https://tools.hmcts.net/jira/browse/PRO-3858

### Change description ###
Update to ensure the formdata.applicantEmail is available prior to creating a case. If the applicantEmail is missing the data is retrieved from the req.session.regId which also holds the applicants email address, although in development environments with no IdAM this value is the sessionId.

```
[ ] Yes
[x] No
```

### DIFF
--- a/app/steps/ui/payment/breakdown/index.js
+++ b/app/steps/ui/payment/breakdown/index.js
@@ -51,12 +51,12 @@ class PaymentBreakdown extends Step {
     * handlePost(ctx, errors, formdata, session, hostname) {
         if (formdata.paymentPending !== 'unknown') {
             if (!formdata.applicantEmail) {
-                logger.warn('Unable to find applicantEmail, using regId ${session.regId} instead.');   
-                if (session.regId !== undefined) {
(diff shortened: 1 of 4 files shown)


app/steps/ui/payment/breakdown/index.js
```diff
@@ -50,6 +50,14 @@ class PaymentBreakdown extends Step {
...
     * handlePost(ctx, errors, formdata, session, hostname) {
         if (formdata.paymentPending !== 'unknown') {
+            if (!formdata.applicantEmail) {
+                logger.warn('Unable to find applicantEmail, using session.regId instead.');
+                if (session.regId) {
+                    formdata.applicantEmail = session.regId;
+                } else {
+                    logger.error('Unable to find applicantEmail or session.regId.');
+                }
+            }
```
shepda: Can we have a unit test to specifically test this new logic, so we can see that if the `formdata.applicantEmail` doesn't exist then it's retrieved from `session.regId`

PR #355, opened 2018-10-05
[PREVIEW] PRO-3858 - Users being prevented from reaching payment screen after probate release

shepda:

https://tools.hmcts.net/jira/browse/PRO-3858

### Change description ###

We need to do this because the formdata gets set without the applicantEmail during the screener questions and the applicantEmail needs to be set after returning from IDAM

```
[ ] Yes
[x] No
```


shepda:
Closed because PREVIEW

PR #359, opened 2018-10-11
PRO-3755 - Upload death certificate - Max Files

shepda:

https://tools.hmcts.net/jira/browse/PRO-3755

### Change description ###

**Add validation for the number of files.**

- Only 10 files may be uploaded and this should work when using the files selector and dragging/dropping.

- If you have multiple error messages displaying on the file list and you remove all the files for an error message then the error summary is also updated to remove reference to that error message.

- If you drag 12 files into the Dropzone and get 2 error messages then remove 2 of the uploaded files, the error messages remain even though you've now only uploaded 8 files...discussed and agreed with Femi and Davy.

```
[ ] Yes
[x] No
```


shepda:
Closed because preview

PR #391, opened 2018-10-25
[PREVIEW] Pro 3779 session not timing out

davejones74:
https://tools.hmcts.net/jira/browse/PRO-3779

### Change description ###
Update to add session expiry to the Probate Personal Applicant application. This is controlled by two config parameters - _Expires_ and _TTL_.
- app.session.expires: 7200000 // ms (120 min)
- app.session.ttl: 28800 // s (8 hours)

The **expires** paremeter configures an inactive period which is two hours. If the user hasn't clicked in the two hours then the application times out. This is configured for two hours to cover the Payment timeout period allowed which is 90 minutes. 

The **ttl** parameter configures the time to live for the session (Redis sessionStore) object. This time is 8 hours but appears to be reset every time the user performs a request.

The difference between ttl and expires is that expires is a soft flag where the session isn't necessirly lost but the application can then force the timeout to occur. This allows CSRF issues to be accomodated. ttl once passed removes the session from redis.

The implementation has highlighted an existing issue currently inherient in the live application and with this update - if the session store object is lost I.e. TTL is reached or the session is removed from Redis unexpectedly, performing a POST at that point would cause an CSRF error, however for a session expiry this is not the case.

Once either a session has expired or a session is lost the application redirects to a /time-out page which is publically and therefore not protected from IdAM. The user will then be given the option to login again. 

```
[ ] Yes
[x] No
```


app/components/security.js
```diff
@@ -32,8 +32,24 @@ class Security {
...
+                    }
+                    logger.info('Redirecting user to the time-out page.');
+                    res.clearCookie(SECURITY_COOKIE);
+                    delete req.cookies[SECURITY_COOKIE];
+                    return res.redirect('/time-out');
+                }
+
+                logger.debug('Extending session for active user.');
+                req.session.expires = Date.now() + config.app.session.expires;
+
```
shepda: We already have the logger on the request and this adds the session id which will help track logging for a specific user, for example use 

`req.log.error('The current user session is lost.');` 

instead of 

`logger.error('The current user session is lost.');`

This will display a log message like this

`2018-11-06T13:55:17+00:00 - error: [frontend, sessionId: ZGy_k7uHQ_9TmMHTdQ6vbGXx69-TtfnQ] Document upload timed out`

instead of 

`2018-11-06T13:55:17+00:00 - error: [frontend, sessionId: Init] Document upload timed out`

PR #429, opened 2018-11-07
PRO-1441: DoB and DoD validation

douglasrice:
https://tools.hmcts.net/jira/browse/PRO-1441

### Change description ###
Validate DoB and DoD relative to each other and only allow dates in the past.

```
[ ] Yes
[X] No
```


app/steps/ui/deceased/dod/index.js
```diff
@@ -14,7 +15,25 @@ class DeceasedDod extends DateStep {
...
+        }
+
+        const dod = new Date(`${ctx.dod_year}-${ctx.dod_month}-${ctx.dod_day}`);
+
+        const today = new Date();
+        today.setHours(0, 0, 0, 0);
+
+        if (dod > today) {
+            errors.push(FieldError('dod_date', 'dateInFuture', this.resourcePath, this.generateContent()));
+        } else if (typeof dob === 'object' && dob >= dod) {
```
shepda: 1. Do we need to check if `dob` is an object here? If so then we'd probably also need to check that `dod` is also an object?

2. Can you add a unit test for this error (`dob >= dod`)?

LucaDelBuonoHMCTS: 1. I'm checking it's an object as it only gets assigned a date if it exists in session.form and I want that line inside the if to get executed only under that circumstance.
2. Weird, I was 100% sure I had added that :-/ I'll add it again

LucaDelBuonoHMCTS: 2. Added back (still sure I did this already, weird)

PR #440, opened 2018-11-09
PRO-3555: Thank You page

LucaDelBuonoHMCTS:
https://tools.hmcts.net/jira/browse/PRO-3555

### Change description ###
Amends to the Thank You page

```
[ ] Yes
[X] No
```


test/component/testThankyouPage.js
```diff
@@ -15,19 +15,22 @@ describe('thank-you', () => {
...
                     nameAsOnTheWill: 'Yes'
                 },
+                ccdCase: {
+                    id: '1234-5678-9012-3456'
+                }
             };
             testWrapper.agent.post('/prepare-session/form')
                 .send(sessionData)
                 .end(() => {
                     const excludeKeys = ['stopParagraph1'];
```
shepda: As `stopParagraph1` has been removed from the content, we can remove it from being excluded in this test

test/component/testThankyouPage.js
```diff
@@ -15,19 +15,22 @@ describe('thank-you', () => {
     });
 
     describe('Verify Content, Errors and Redirection', () => {
-
         it('test right content loaded on the page when NO soft stop', (done) => {
             const sessionData = {
                 applicant: {
                     nameAsOnTheWill: 'Yes'
                 },
```
shepda: We're not doing using `applicant.nameAsOnTheWill` on this page so we probably don't need a test for soft stop and no soft stop and these two tests are essentially testing the same thing.

app/steps/ui/thankyou/index.js
```diff
@@ -8,6 +8,14 @@ class ThankYou extends Step {
         return '/thankyou';
     }
 
+    getContextData(req) {
+        const ctx = super.getContextData(req);
+        if (req.session.form.ccdCase && req.session.form.ccdCase.id) {
+            ctx.ccdReferenceNumber = req.session.form.ccdCase.id;
+        }
+        return ctx;
+    }
```
shepda: Please add unit tests for this method

app/steps/ui/thankyou/index.js
```diff
@@ -8,9 +8,12 @@ class ThankYou extends Step {
...
-    handleGet(ctx, formdata) {
-        ctx.softStop = this.anySoftStops(formdata, ctx);
-        return [ctx];
+    getContextData(req) {
+        const ctx = super.getContextData(req);
+        if (req.session.form.ccdCase && req.session.form.ccdCase.id) {
+            ctx.ccdReferenceNumber = req.session.form.ccdCase.id;
+        }
+        return ctx;
     }
```
shepda: `getContextData()` gets called for both `handleGet()` and `handlePost()` so we'll need to remove `ctx.ccdReferenceNumber` in `action()` otherwise it'll get persisted when called by `handlePost()`.

Please also add unit tests for `action()`

app/steps/ui/thankyou/index.js
```diff
@@ -10,6 +10,7 @@ class ThankYou extends Step {
 
     getContextData(req) {
         const ctx = super.getContextData(req);
+        ctx.ccdReferenceNumber = '';
```
shepda: Can we have a unit test to check that `ctx.ccdReferenceNumber` is set to an empty string if the case id doesn't exist?

app/steps/ui/thankyou/template.html
```diff
@@ -10,10 +10,12 @@
...
-                    <p class="font-large">
-                        <span class="font-medium">{{content.referenceNumber | safe}}</span><br>
-                        <strong class="bold">{{fields.ccdReferenceNumber.value}}</strong>
-                    </p>
+                    {% if fields.ccdReferenceNumber.value !== '' %}
+                        <p class="font-large">
+                            <span class="font-medium">{{content.referenceNumber | safe}}</span><br>
+                            <strong class="bold">{{fields.ccdReferenceNumber.value}}</strong>
+                        </p>
+                    {% endif %}
```
shepda: Can we have a component test to check that this content is not displayed if `fields.ccdReferenceNumber.value` is an empty string?

test/component/testThankYouPage.js
```diff
@@ -15,7 +15,23 @@ describe('thank-you', () => {
...
+                .send(sessionData)
+                .end(() => {
+                    const excludeKeys = ['referenceNumber'];
+
+                    const contentData = {
+                        helpLineNumber: config.helpline.number,
+                        findOutNext: config.links.findOutNext
+                    };
+
+                    testWrapper.testContent(done, excludeKeys, contentData);
```
shepda: It would be better to use `testWrapper.testContentNotPresent()` in this test because excluding the content will make the test pass by simply not checking that it exists in the page.

For example, if you excluded the same content in the `test content loaded on the page when CCD Case ID present` test, where this content should appear, then the test would still pass.

app/steps/ui/thankyou/index.js
```diff
@@ -12,9 +12,9 @@ class ThankYou extends Step {
         const ctx = super.getContextData(req);
         ctx.ccdReferenceNumber = '';
         if (req.session.form.ccdCase && req.session.form.ccdCase.id) {
-            ctx.ccdReferenceNumber = req.session.form.ccdCase.id;
+            ctx.ccdReferenceNumber = req.session.form.ccdCase.id.toString();
```
shepda: Can we add a unit test for this so that we check that we get the expected output if `req.session.form.ccdCase.id` is passed as an integer.

It could be a similar test to `should return the context with the CCD Case ID when present (WITHOUT dashes)` in `test/unit/testThankYouPage.js` but providing an integer instead of a string.

PR #446, opened 2018-11-12
Pro 2795 remove escaped characters in declarartion

sanjayparekh-hmcts:
Remove escaped chars for print PA template

```
[ ] Yes
[ X] No
```


app/resources/en/translation/declaration.json
```diff
@@ -1,6 +1,6 @@
 {
   "title": "Check the legal statement and make your declaration",
-  "intro": "This statement is based on the information you&rsquo;ve given in your application. It will be stored as a public record.",
+  "intro": "This statement is based on the information you've given in your application. It will be stored as a public record.",
```
shepda: We should be using `&rsquo;` here to keep consistency across the app

sanjayparekh-hmcts: I dont think we should have escaped chars in 'content' if we can avoid it

shepda: We've done this throughout the app because the decision was taken a long time ago by UX that having proper 66 and 99's was looked better than having `'`.

If we're going to decide to change that then we should change them throughout the app rather than just in one place.

sanjayparekh-hmcts: I've had to reinstate these as this is the only way t solve this JIRA

PR #447, opened 2018-11-12
Pro 3779 session not timing out

davejones74:
https://tools.hmcts.net/jira/browse/PRO-3779

### Change description ###
Update to add session expiry to the Probate Personal Applicant application. This is controlled by two config parameters - _Expires_ and _TTL_.
- app.session.expires: 3600000 // ms (60 min)
- app.session.ttl: 28800 // s (8 hours)

The **expires** paremeter configures an inactive period which is 60 minutes. If the user hasn't clicked in the 60 minutes then the application times out. There is an exclude list **whitelistedPagesIgnoreSessionTimeout** which allows the /payment-status page to be ignored during payments so these don't expire. 

The **ttl** parameter configures the time to live for the session (Redis sessionStore) object. This time is 8 hours but appears to be reset every time the user performs a request.

The difference between ttl and expires is that expires is a soft flag where the session isn't necessirly lost but the application can then force the timeout to occur. This allows CSRF issues to be accomodated. ttl once passed removes the session from redis.

The implementation has highlighted an existing issue currently inherient in the live application and with this update - if the session store object is lost I.e. TTL is reached or the session is removed from Redis unexpectedly, performing a POST at that point would cause an CSRF error, however for a session expiry this is not the case.

Once either a session has expired or a session is lost the application redirects to a /time-out page which is publically and therefore not protected from IdAM. The user will then be given the option to login again. 

```
[ ] Yes
[x] No
```


app/components/security.js
```diff
@@ -35,10 +36,11 @@ class Security {
             if (securityCookie) {
                 const lostSession = !req.session.expires;
                 const sessionExpired = req.session.expires?req.session.expires <= Date.now():false;
+                const sessionTimeoutCheck = !includes(config.whitelistedPagesIgnoreSessionTimeout, req.originalUrl);
```
LucaDelBuonoHMCTS: Do we need lodash for this? Wouldn't `!config.whitelistedPagesIgnoreSessionTimeout.includes(req.originalUrl)` be the same?

app/components/security.js
```diff
@@ -76,7 +98,7 @@ class Security {
         if (userRoles.some(role => authorisedRoles.includes(role))) {
             next();
         } else {
-            logger.error('[ERROR] :: Error authorising user, Role Not Authorised');
+            logger().error('[ERROR] :: Error authorising user, Role Not Authorised');
```
LucaDelBuonoHMCTS: Why not use `req.log.error()` here like above?

davejones74: req not passed in so can't create req.log from logger(req.sessionID).

app/config.js
```diff
@@ -163,7 +164,7 @@ const config = {
...
     whitelistedPagesAfterDeclaration: ['/tasklist', '/executors-invites-sent', '/copies-uk', '/assets-overseas', '/copies-overseas', '/copies-summary', '/payment-breakdown', '/payment-breakdown?status=failure', '/payment-status', '/documents', '/thankyou', '/check-answers-pdf', '/declaration-pdf', '/sign-out'],
     hardStopParams: ['will.left', 'will.original', 'iht.completed', 'applicant.executor'],
-    nonIdamPages: ['stop-page/*', 'error', 'sign-in', 'pin-resend', 'pin-sent', 'co-applicant-*', 'pin', 'inviteIdList', 'start-eligibility', 'death-certificate', 'deceased-domicile', 'iht-completed', 'will-left', 'will-original', 'applicant-executor', 'mental-capacity', 'died-after-october-2014', 'relationship-to-deceased', 'other-applicants', 'start-apply', 'time-out'],
+    nonIdamPages: ['stop-page/*', 'error', 'sign-in', 'pin-resend', 'pin-sent', 'co-applicant-*', 'pin', 'inviteIdList', 'start-eligibility', 'death-certificate', 'deceased-domicile', 'iht-completed', 'will-left', 'will-original', 'applicant-executor', 'mental-capacity', 'died-after-october-2014', 'relationship-to-deceased', 'other-applicants', 'time-out', 'start-apply'],
```
LucaDelBuonoHMCTS: Pointless change :)

app/config.js
```diff
@@ -173,8 +174,11 @@ const config = {
...
-    hardStopParams: ['will.left', 'will.original', 'iht.completed', 'applicant.executor'],
-    nonIdamPages: ['stop-page/*', 'error', 'sign-in', 'pin-resend', 'pin-sent', 'co-applicant-*', 'pin', 'inviteIdList', 'start-eligibility', 'death-certificate', 'deceased-domicile', 'iht-completed', 'will-left', 'will-original', 'applicant-executor', 'mental-capacity', 'died-after-october-2014', 'relationship-to-deceased', 'other-applicants', 'start-apply', 'time-out'],
+    hardStopParams: {
+        probate: [],
+        intestacy: []
+    },
+    nonIdamPages: ['stop-page/*', 'error', 'sign-in', 'pin-resend', 'pin-sent', 'co-applicant-*', 'pin', 'inviteIdList', 'start-eligibility', 'death-certificate', 'deceased-domicile', 'iht-completed', 'will-left', 'will-original', 'applicant-executor', 'mental-capacity', 'died-after-october-2014', 'related-to-deceased', 'other-applicants', 'start-apply', 'time-out'],
```
LucaDelBuonoHMCTS: Weird that this block shows as a change with master... :-/

bmj-vsalivendra:
Closing this pr and creating new one because Unable to access feature toggles.

PR #460, opened 2018-11-19
PRI-32: Screening Questions Redirects

LucaDelBuonoHMCTS:
https://tools.hmcts.net/jira/browse/PRI-32
(previously https://tools.hmcts.net/jira/browse/PRO-3082)

### Change description ###
Forces the user to answer all the screening questions by setting a cookie and keeping track of what questions have been asked

```
[ ] Yes
[X] No
```


test/util/TestWrapper.js
```diff
@@ -162,6 +174,22 @@ class TestWrapper {
...
+                const cookieName = cookies[i].name;
+                const cookieContent = JSON.stringify(cookies[i].content);
+                cookiesString = `${cookieName}=${cookieContent},`;
+            }
+
+            cookiesString = cookiesString.substring(0, cookiesString.length - 1);
+
+            res.set('Cookie', cookiesString);
+        }
+    }
```
shepda: It would be better for this method to return the cookie string and then to set that to `res` in the calling method.

This way `setCookies()` has predictable input and outputs.

app/utils/EligibilityCookie.js
```diff
@@ -0,0 +1,58 @@
...
+
+        if (req.protocol === 'https') {
+            res.cookie(ELIGIBILITY_COOKIE, cookieValue, {secure: true, httpOnly: true});
+        } else {
+            res.cookie(ELIGIBILITY_COOKIE, cookieValue, {httpOnly: true});
+        }
+    }
+}
+
+module.exports = EligibilityCookie;
```
shepda: Please add unit tests for all the new methods in this class

PR #462, opened 2018-11-22
PRO-3484: Amends to the Documents page

LucaDelBuonoHMCTS:
https://tools.hmcts.net/jira/browse/PRO-3484

### Change description ###
Copy amends to Documents page behind Document Upload feature toggle

```
[ ] Yes
[X] No
```


app/steps/ui/documents/index.js
```diff
@@ -5,27 +5,46 @@ const ExecutorsWrapper = require('app/wrappers/Executors');
...
+        delete ctx.hasCodicils;
+        delete ctx.codicilsNumber;
+        delete ctx.hasMultipleApplicants;
+        delete ctx.hasRenunciated;
+        delete ctx.is205;
+        delete ctx.executorsNameChangedByDeedPollList;
+        delete ctx.ccdReferenceNumber;
+        delete ctx.isDocumentUploadToggleEnabled;
+        return [ctx, formdata];
+    }
```
shepda: We don't need to define `action()`  in this class to delete the `ctx` variables as they're all defined in `handleGet()` and this method doesn't persist to the database.

LucaDelBuonoHMCTS: Weird, I don't remember adding these lines... removed now.

test/unit/testDocumentsPage.js
```diff
@@ -14,4 +13,22 @@ describe('Documents', () => {
...
+                hasRenunciated: false,
+                is205: true,
+                executorsNameChangedByDeedPollList: [],
+                ccdReferenceNumber: '1234-1235-1236-1237',
+                isDocumentUploadToggleEnabled: true
+            };
+            [ctx] = Documents.action(ctx);
+            assert.deepEqual(ctx, {});
+        });
+    });
```
shepda: Please remove these redundant tests

test/component/testDocumentsPage.js
```diff
@@ -1190,6 +1190,7 @@ describe('documents', () => {
                         'old_sendDocumentsAddress',
                         'text6',
                         'checklist1-item1-codicils',
+                        'checklist2-item3-will-uploaded',
```
shepda: This test is testing that the content is not displayed if the user has uploaded a document. 

We also need a similar test to check that the content is displayed if the user has not uploaded a document...uploading a document is not a requirement.

PR #475, opened 2018-12-03
Pri 24 remove screener data submission test submit service

bmj-vsalivendra:
PRI-24: remove screener questions
PRI-91 : pointe to submit service
Does this PR introduce a breaking change? (check one with "x")

[ ] Yes
[ X] No


test/service-stubs/business.js
```diff
@@ -1,10 +1,12 @@
 'use strict';
 
 /* eslint no-console: 0 */
```
shepda: Please remove this eslint override as we've removed the `console.log()`

commit-msg.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 COMMIT_MESSAGE=`head -n1 $1`
-PATTERN="^(PRO|PRI)-[0-9]+: "
+PATTERN="^PRO-[0-9]+: |^PRI-[0-9]+: "
```
shepda: I think we'd be better to use the original regex pattern as its clearer and has less duplication.

app/components/services.js
```diff
@@ -21,6 +21,7 @@ const SERVICE_AUTHORISATION_URL = `${config.services.idam.s2s_url}/lease`;
 const serviceName = config.services.idam.service_name;
 const secret = config.services.idam.service_key;
 const FEATURE_TOGGLE_URL = config.featureToggles.url;
+const FEATURE_TOGGLE_PORT = config.featureToggles.port;
```
shepda: We need the port for the feature toggles locally but we don't have one in the AKS envs, so by splitting out the port number it means that the app will use the default value locally and it will also use it in the AKS envs, because we won't have a value.

We have a `FormatUrl` util, which will use the port number if we have one and ignore it (including the `:`) if we don't have one but we cannot use that in this case because we'll always have the default value.

I think for now it would be better to keep it as it was with feature toggles in the persistence stub.

sanjayparekh-hmcts: Are you sure? we talked about this yesterday + you said it was OK

shepda: Yep, I thought so but after speaking to Praveen about it today, it looks like it's not going to work in all envs because we only have the port number locally...which works when it's in the url but not when it's split out to a separate param.

PR #479, opened 2018-12-04
PRI-110: Set expiry date on eligibility cookie to 2 days

LucaDelBuonoHMCTS:
https://tools.hmcts.net/jira/browse/PRI-110

### Change description ###
Sets expiry date on eligibility cookie to 2 days

```
[ ] Yes
[X] No
```


app/utils/EligibilityCookie.js
```diff
@@ -50,9 +50,9 @@ class EligibilityCookie {
         const cookieValue = JSON.stringify(json);
 
         if (req.protocol === 'https') {
-            res.cookie(eligibilityCookieName, cookieValue, {httpOnly: true, secure: true, expires: new Date(Date.now() + 172800)});
+            res.cookie(eligibilityCookieName, cookieValue, {httpOnly: true, expires: new Date(Date.now() + 172800), secure: true});
         } else {
-            res.cookie(eligibilityCookieName, cookieValue, {httpOnly: true});
+            res.cookie(eligibilityCookieName, cookieValue, {httpOnly: true, expires: new Date(Date.now() + 172800)});
```
shepda: I think it would be better to define the expiry time (`172800`) in the config and use it from there in all the places.

Also we could reduce the code complexity of this conditional by creating the cookie params object before the conditional, setting `secure: true` only in the conditional and creating the cookie after the conditional

LucaDelBuonoHMCTS: Done

app/utils/EligibilityCookie.js
```diff
@@ -27,17 +27,9 @@ class EligibilityCookie {
...
-        }
-
-        const fieldFound = Object.keys(json.pages[currentPage]).includes(fieldKey);
-
-        if (!fieldFound) {
-            json.pages[currentPage][fieldKey] = fieldValue;
-        }
+        json.nextStepUrl = nextStepUrl;
+        json.pages[currentPage] = {};
+        json.pages[currentPage][fieldKey] = fieldValue;
```
shepda: I think we can simplify the unit tests for `setCookie()` with this change

PR #499, opened 2018-12-10
PRO-4001: add document download to thankyou page

Wivenhoe-HMCTS:
Add document download for pdfs to thankyou page

```
[ ] Yes
[ x] No
```

### DIFF
--- a/app/steps/ui/thankyou/index.js
+++ b/app/steps/ui/thankyou/index.js
@@ -8,6 +8,21 @@ class ThankYou extends Step {
         return '/thankyou';
     }
 
+    handleGet(ctx, formdata) {
+        ctx.displaySaveApplication = false;
+        ctx.checkAnswersSummary = false;
+        ctx.legalDeclaration = false;
+        if (formdata.checkAnswersSummary) {
+            ctx.checkAnswersSummary = true;
+            ctx.displaySaveApplication = true;
+        }
+        if (formdata.legalDeclaration) {
+            ctx.legalDeclaration = true;
+            ctx.displaySaveApplication = true;
+        }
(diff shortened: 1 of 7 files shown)


app/steps/ui/thankyou/index.js
```diff
@@ -8,6 +8,21 @@ class ThankYou extends Step {
...
+        if (formdata.checkAnswersSummary) {
+            ctx.checkAnswersSummary = true;
+            ctx.displaySaveApplication = true;
+        }
+        if (formdata.legalDeclaration) {
+            ctx.legalDeclaration = true;
+            ctx.displaySaveApplication = true;
+        }
+        return [ctx];
+    }
```
shepda: Do we need to know what value `checkAnswersSummary` and `legalDeclaration` have after the user has finished the application? For example if they log back in again or for a backend service?

If not and this value is just used for the current session then it would be better to store in the session rather than in the form data and therefore persisting it to the db.

Wivenhoe-HMCTS: Yes the user can log back in and will directed to this thank you page. The conditions around the checkAnswersSummary & legalDeclaration allow for the hyperlinks to be displayed only if the document still exits within the form data. 
Also we don't need to data migrate any historic probate applications that may have not created the pdf files originally. They will simply see the default thank you page without any references to the pdf's. 
There is another follow on jira that may store the pdf's into the evidence management system so that we can keep  the pdfs for the backoffice processes as well.

Wivenhoe-HMCTS:
closing PR , since jenkins pipeline is failing and Praveen says a new PR is required

PR #502, opened 2018-12-12
Pro 4001 add pdf hyperlinks to thankyou page

Wivenhoe-HMCTS:
add document download links to thankyou page 

```
[ ] Yes
[x ] No
```

### DIFF
--- a/test/unit/testThankYouPage.js
+++ b/test/unit/testThankYouPage.js
@@ -46,4 +46,49 @@ describe('ThankYou', () => {
             expect(ctx).to.deep.equal({});
         });
     });
+
+    describe('handleGet()', () => {
+        it('test when only checkAnswersSummary JSON exists', () => {
+            let ctx = {};
+            let formdata = {
+                checkAnswersSummary: '{"test":"data"}'
+            };
+            [ctx, formdata] = ThankYou.handleGet(ctx, formdata);
+            expect(ctx.checkAnswersSummary).to.deep.equal(true);
+            expect(ctx.legalDeclaration).to.deep.equal(false);
+            expect(ctx.displaySaveApplication).to.deep.equal(true);
+        });
+
+        it('test when only legalDeclaration JSON exists', () => {
+            let ctx = {};
+            let formdata = {
+                legalDeclaration: '{"test":"data"}'
+            };
+            [ctx, formdata] = ThankYou.handleGet(ctx, formdata);
+            expect(ctx.checkAnswersSummary).to.deep.equal(false);
+            expect(ctx.legalDeclaration).to.deep.equal(true);
+            expect(ctx.displaySaveApplication).to.deep.equal(true);
+        });
+
+        it('test when no pdf variables JSON exists', () => {
+            let ctx = {};
+            let formdata = {};
+            [ctx, formdata] = ThankYou.handleGet(ctx, formdata);
+            expect(ctx.checkAnswersSummary).to.deep.equal(false);
+            expect(ctx.legalDeclaration).to.deep.equal(false);
+            expect(ctx.displaySaveApplication).to.deep.equal(false);
+        });
+
+        it('test when all pdf variables JSON exists', () => {
+            let ctx = {};
+            let formdata = {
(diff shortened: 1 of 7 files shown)


test/unit/testThankYouPage.js
```diff
@@ -46,4 +46,49 @@ describe('ThankYou', () => {
...
+        it('test when only checkAnswersSummary JSON exists', () => {
+            let ctx = {};
+            let formdata = {
+                checkAnswersSummary: '{"test":"data"}'
+            };
+            [ctx, formdata] = ThankYou.handleGet(ctx, formdata);
+            expect(ctx.checkAnswersSummary).to.deep.equal(true);
+            expect(ctx.legalDeclaration).to.deep.equal(false);
+            expect(ctx.displaySaveApplication).to.deep.equal(true);
+        });
```
shepda: `formdata` isn't required to be set on line 56 as `handleGet()` is only returning `[ctx]`.

The expect lines don't need to use `expect().to.deep.equal()` as they're testing for a single value rather than an array/object so you could use `expect().to.equal()`.

Rather than having an assertion for each of the `ctx` values, it may be more readable to have a single assertion for the entire `ctx` value.

test/unit/testThankYouPage.js
```diff
@@ -46,4 +46,49 @@ describe('ThankYou', () => {
...
+        it('test when only legalDeclaration JSON exists', () => {
+            let ctx = {};
+            let formdata = {
+                legalDeclaration: '{"test":"data"}'
+            };
+            [ctx, formdata] = ThankYou.handleGet(ctx, formdata);
+            expect(ctx.checkAnswersSummary).to.deep.equal(false);
+            expect(ctx.legalDeclaration).to.deep.equal(true);
+            expect(ctx.displaySaveApplication).to.deep.equal(true);
+        });
```
shepda: `formdata` isn't required to be set on line 67 as `handleGet()` is only returning `[ctx]`.

The expect lines don't need to use `expect().to.deep.equal()` as they're testing for a single value rather than an array/object so you could use `expect().to.equal()`.

Rather than having an assertion for each of the `ctx` values, it may be more readable to have a single assertion for the entire `ctx` value.

test/unit/testThankYouPage.js
```diff
@@ -46,4 +46,49 @@ describe('ThankYou', () => {
...
+
+        it('test when no pdf variables JSON exists', () => {
+            let ctx = {};
+            let formdata = {};
+            [ctx, formdata] = ThankYou.handleGet(ctx, formdata);
+            expect(ctx.checkAnswersSummary).to.deep.equal(false);
+            expect(ctx.legalDeclaration).to.deep.equal(false);
+            expect(ctx.displaySaveApplication).to.deep.equal(false);
+        });
```
shepda: `formdata` isn't required to be set on line 76 as `handleGet()` is only returning `[ctx]`.

The expect lines don't need to use `expect().to.deep.equal()` as they're testing for a single value rather than an array/object so you could use `expect().to.equal()`.

Rather than having an assertion for each of the `ctx` values, it may be more readable to have a single assertion for the entire `ctx` value.

test/unit/testThankYouPage.js
```diff
@@ -46,4 +46,49 @@ describe('ThankYou', () => {
...
+            let formdata = {
+                checkAnswersSummary: '{"test":"data"}',
+                legalDeclaration: '{"test":"data"}'
+            };
+            [ctx, formdata] = ThankYou.handleGet(ctx, formdata);
+            expect(ctx.checkAnswersSummary).to.deep.equal(true);
+            expect(ctx.legalDeclaration).to.deep.equal(true);
+            expect(ctx.displaySaveApplication).to.deep.equal(true);
+        });
+    });
```
shepda: `formdata` isn't required to be set on line 88 as `handleGet()` is only returning `[ctx]`.

The expect lines don't need to use `expect().to.deep.equal()` as they're testing for a single value rather than an array/object so you could use `expect().to.equal()`.

Rather than having an assertion for each of the `ctx` values, it may be more readable to have a single assertion for the entire `ctx` value.

PR #523, opened 2018-12-21
PRO-4392: Tasklist Bug with Executors Roles

LucaDelBuonoHMCTS:
https://tools.hmcts.net/jira/browse/PRO-4392

### Change description ###
Fix bug with Executor Roles question being marked as complete when an executor renounces but Executor Notified question still not answered, which leads to Tasklist marking Executors task as completed

```
[ ] Yes
[X] No
```


app/steps/ui/executors/roles/index.js
```diff
@@ -40,7 +40,16 @@ class ExecutorRoles extends CollectionStep {
...
+        return [every(ctx.list, exec => {
+            return exec.isApplying ||
+                (
+                    !isEmpty(exec.notApplyingReason) &&
+                    (
+                        (exec.notApplyingReason === json.optionPowerReserved && !isEmpty(exec.executorNotified)) ||
+                        (exec.notApplyingReason !== json.optionPowerReserved)
+                    )
+                );
+        }), 'inProgress'];
```
shepda: Please add unit tests for this logic

LucaDelBuonoHMCTS: Added

test/unit/testExecutorRoles.js
```diff
@@ -63,6 +63,132 @@ describe('ExecutorRoles', () => {
         });
     });
 
+    describe('isComplete()', () => {
+        it('should return the correct step completion status status when all executors are applying', (done) => {
```
shepda: Please remove the duplicated 'status' from the description in these tests

test/unit/testExecutorRoles.js
```diff
@@ -63,6 +63,132 @@ describe('ExecutorRoles', () => {
...
+            const ctx = {
+                'list': [
```
shepda: Please remove the quotes from object params in these tests to keep consistency throughout the file

LucaDelBuonoHMCTS: Ha good spot, I actually pasted it from some other file so will search for that and remove the quotes from there as well.

LucaDelBuonoHMCTS: I found the same all over the place... removing them all.

PR #562, opened 2019-01-22
PRI-114 - Create intestacy config/services

shepda:

https://tools.hmcts.net/jira/browse/PRI-114

### Change description ###

Refactor services from `app/components/services.js` into clients which can use used for both probate and intestacy journeys and also the pact tests.

```
[ ] Yes
[x] No
```


reddy1praveen:
https://github.com/hmcts/probate-frontend/pull/570 
opened here

PR #572, opened 2019-02-01
Charts Feature Toggle & Redis

damongreen123:

### Change description ###

```
[ ] Yes
[ ] No
```


charts/probate-frontend/values.template.yaml
```diff
@@ -0,0 +1,416 @@
...
+    securityContext:
+      enabled: true
+      fsGroup: 1001
+      runAsUser: 1001
+
+    ## Enable persistence using Persistent Volume Claims
+    ## ref: http://kubernetes.io/docs/user-guide/persistent-volumes/
+    ##
+    persistence:
+      enabled: true
```
timwebster9: Is this really needed? It creates a Persistent Volume (i.e. disk) in the cluster.  It will increase provisioning time, increase costs and add complexity.  Suggest you disable it unless for some reason you really need it - the local container storage should be fine for a PR build shouldn't it?

damongreen123: Cheers for pointing that out Tim, I will change it. I just threw the whole redis default config in there as a starting point. Probably also need to review security options.

charts/probate-frontend/values.template.yaml
```diff
@@ -0,0 +1,416 @@
+nodejs:
+  applicationPort: 3000
+  cpuRequests: 0.2
```
timwebster9: Can you also remove this? The default of `25m` is fine.  We are trying to pack the nodes a little denser..

damongreen123: Think I pinched that from cmc-citizen-frontend. Ive removed.

damongreen123:
opened a cleaner branch

PR #594, opened 2019-02-18
PRO-4264: Dealing with estate prune formdata

Bishwin:
https://tools.hmcts.net/jira/browse/PRO-4264

### Change description ###
This PR cleans up redundant formdata when an executor is no longer dealing with the estate.

This PR also fixes a bug found on executors-additional-invite. Steps to reproduce the issue on master are below
```
add 2 executors
all alive
all dealing with estate
no alias
notify executors
sign out and in
change executors to 4
enter information
sign declaration
taken to additional-invite page
view urls for amending details - indexes should be off
```

```
[ ] Yes
[ ] No
```

### DIFF
--- a/app/steps/ui/executors/additionalinvite/index.js
+++ b/app/steps/ui/executors/additionalinvite/index.js
@@ -13,6 +13,7 @@ class ExecutorsAdditionalInvite extends ValidationStep {
 
     getContextData(req) {
         const ctx = super.getContextData(req);
+        ctx.executorsToNotifyList = new ExecutorsWrapper(ctx).executorsToNotify();
         ctx.inviteSuffix = size(ctx.executorsToNotifyList) > 1 ? '-multiple' : '';
         ctx.executorsToNotifyNames = FormatName.formatExecutorNames(ctx.executorsToNotifyList);
         return ctx;
--- a/app/steps/ui/executors/contactdetails/index.js
+++ b/app/steps/ui/executors/contactdetails/index.js
@@ -3,6 +3,7 @@
 const ValidationStep = require('app/core/steps/ValidationStep');
 const emailValidator = require('email-validator');
 const validator = require('validator');
+const ExecutorsWrapper = require('app/wrappers/Executors');
 const FieldError = require('app/components/error');
 const {findIndex, every, tail} = require('lodash');
 const InviteData = require('app/services/InviteData');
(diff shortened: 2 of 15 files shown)


app/steps/ui/executors/contactdetails/index.js
```diff
@@ -50,8 +48,6 @@ class ExecutorContactDetails extends ValidationStep {
             executor.emailChanged = true;
```
davejones74: As per our conversation during a test run when changing the Executor email (after notifying) I wasn't able to then renotify the executor. The notification screen didn't show and the "Check your answers and make your legal declaration" was still selectable so the user would end up going around in a loop.

SamanthaSJ:
Verified PR and ticket 4264 updated. Pass.

app/steps/ui/executors/additionalinvite/index.js
```diff
@@ -13,6 +13,7 @@ class ExecutorsAdditionalInvite extends ValidationStep {
 
     getContextData(req) {
         const ctx = super.getContextData(req);
+        ctx.executorsToNotifyList = new ExecutorsWrapper(ctx).executorsToNotify();
```
LucaDelBuonoHMCTS: Isn't this redundant?
`executorsToNotifyList` should already be in `session.form.data.executors`, therefore `const ctx = super.getContextData(req);` should add it to `ctx`.

test/component/executors/testExecutorsAdditionalInvite.js
```diff
@@ -48,7 +48,7 @@ describe('executors-additional-invite', () => {
         });
 
         it('test content displays only the executors who have been added and need to be emailed', (done) => {
-            sessionData.executors.executorsToNotifyList = [
+            sessionData.executors.list = [
```
LucaDelBuonoHMCTS: In case what I commented above is correct, then this is not necessary either.

test/unit/testExecutorsAdditionalInvite.js
```diff
@@ -25,51 +25,31 @@ describe('Executor-Additional-Invite', function () {
 
         it('test that context variables are correctly populated for single executor to be notified', () => {
             req.session.form.executors = {
-                executorsToNotifyList: [
+                list: [
```
LucaDelBuonoHMCTS: Same as above.

PR #704, opened 2019-04-11
Pro 4325 check payment status using ccd reference

davejones74:
https://tools.hmcts.net/jira/browse/PRO-4325

### Change description ###
Update payment-breakdown to check payment status by looking at the payments using a case id. If any Initiated or Success payments are found and different from the currently tracked paymentid then this is processed in payment-status instead. 

The advantages of this are:
- Handles payment references being out of sync with the frontend application.
- Protects againast duplicate payments (serverside - still requires client check).
- Allows the callback from payments to be detected within the citizen user journey.

*Now includes merged changes from PRO-5109 which has been independantly dev checked.*

```
[ ] Yes
[x] No
```


app/services/Payment.js
```diff
@@ -16,19 +17,47 @@ class Payment extends Service {
...
     }
+
+    identifySuccessfulOrInitiatedPayment(casePayments) {
+        for (const payment of casePayments.payments) {
+            if (payment.status === 'Success') {
+                this.log(`Found successful payment: ${payment.payment_reference}`);
+                return payment;
+            }
+        }
+        for (const payment of casePayments.payments) {
```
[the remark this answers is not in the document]
davejones74: This is because when checking the payments for a case, if I find a Success then I can ignore any Initiated states. I can only do this if run through the entire array of payments first.

app/steps/ui/payment/breakdown/index.js
```diff
@@ -66,14 +66,33 @@ class PaymentBreakdown extends Step {
...
+            const keyword = 'failure';
+            errors.push(FieldError('authorisation', keyword, this.resourcePath, ctx));
+            return [ctx, errors];
         }
-        const canCreatePayment = yield this.canCreatePayment(ctx, formdata, serviceAuthResult);
-        if (formdata.paymentPending !== 'unknown') {
+
+        const [canCreatePayment, paymentStatus] = yield this.canCreatePayment(ctx, formdata, serviceAuthResult);
+        logger.info(`canCreatePayment result = ${canCreatePayment} with status ${paymentStatus}`);
+        if (paymentStatus === 'Initiated' && canCreatePayment === false) {
```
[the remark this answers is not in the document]
davejones74: Yes I agree, the canCreatePayment will always be false if paymentStatus is Initiated. I've removed the second condition.

app/services/Payment.js
```diff
@@ -16,19 +17,47 @@ class Payment extends Service {
...
             'ServiceAuthorization': data.serviceAuthToken,
-            'return-url': this.formatUrl.format(hostname, '/payment-status')
+            'return-url': this.formatUrl.format(hostname, config.services.payment.paths.returnUrlPath)
         };
         const body = paymentData.createPaymentData(data);
         const fetchOptions = this.fetchOptions(body, 'POST', headers);
         return [this.fetchJson(url, fetchOptions), body.reference];
     }
+
+    identifySuccessfulOrInitiatedPayment(casePayments) {
```
douglasrice: As this function is really checking if a payment is made it should return true or false.  A function should always return something.  - This will also remove the check for 'undefined' in the calling routine.
Also, we use a library call lodash which has array processing.  To simplify the array processing you could use the following:
_.indexOf(['Initiated',Iinitiated', 'Initiated', 'Success'], 'Success');

app/routes.js
```diff
@@ -66,6 +66,7 @@ router.use((req, res, next) => {
     const isHardStop = (formdata, journey) => config.hardStopParams[journey].some(param => get(formdata, param) === commonContent.no);
     const executorsWrapper = new ExecutorsWrapper(formdata.executors);
     const hasMultipleApplicants = executorsWrapper.hasMultipleApplicants();
+    const skipPayment = get(formdata, 'payment.total') === 0 || get(formdata, 'payment.status') === 'Success';
```
douglasrice: I realise the skipPayment makes reading the if/then/else statement a bit easier however, the if/then/else statement has this type of logic "inline".  We should either keep the code as the rest is or change the others to be like skipPayment - otherwise we have a mismatch of coding practices

app/steps/ui/payment/breakdown/index.js
```diff
@@ -178,23 +182,37 @@ class PaymentBreakdown extends Step {
...
-            logger.info(`Payment retrieval in breakdown for paymentId = ${ctx.paymentId} with response = ${JSON.stringify(paymentResponse)}`);
+            const paymentServiceUrl = config.services.payment.url + config.services.payment.paths.payments;
+            const payment = new Payment(paymentServiceUrl, ctx.sessionID);
+            const casePaymentsArray = yield payment.getCasePayments(data);
+            logger.debug(`Case payments for ${caseId} with response = ${JSON.stringify(casePaymentsArray)}`);
+            const paymentResponse = payment.identifySuccessfulOrInitiatedPayment(casePaymentsArray);
+            logger.debug(`Payment retrieval in breakdown for caseId = ${caseId} with response = ${JSON.stringify(paymentResponse)}`);
             if (typeof paymentResponse === 'undefined') {
-                return true;
+                logger.info('No pyaments of Initiated for Success found for case.');
```
douglasrice: Typo: pyaments

app/services/Payment.js
davejones74: The _identifySuccessfulOrInitiatedPayment_ function is bringing back the paymentDto where status is Success or Initiated. In actual fact it only needs the reference so possibily it could be simplified but I'm not sure Lodash indexOf can support key expressions in JSON? So I'd need to create some temporary arrays?

I've created a let variable for an undefinedPayment and I'm returning this if no payment is found, makes it a little easier to follow instead of not returning any value I guess.

LucaDelBuonoHMCTS: I guess you could have just had a `return false` at the end of the function, no need to have that let variable.
Then in the check you could just have `if (paymentResponse) {`.

PR #745, opened 2019-05-09
PRO-4741: Webchat plugin

LucaDelBuonoHMCTS:


app/views/includes/scripts.html
```diff
@@ -47,3 +47,14 @@
 <!--[if lte IE 9]>
 <script src="/public/javascripts/form-polyfill.js"></script>
 <![endif]-->
+
+<script src="/webchat/javascript/hmcts-webchat.js"></script>
+<script nonce="{{ nonce }}">
+    webchat_init({
+        uuid: 'script_4860291385c640afb2487b9.55856720',
```
reddy1praveen: This is test Keys.
Need to add this in config.js and get it from ENV. variables.

PR #760, opened 2019-05-17
PRO-4080: Updated frontend to point to submit pr for testing 4080.

davejones74:
https://tools.hmcts.net/jira/browse/PRO-4080

### Change description ###
Update to allow a 'system' caseworker user with the role caseworker_probate_issuer to update CCD caveat and grantofrepresentation cases using the Probate Orchestrator service. This is to allow a new endpoint to be defined /update-payments which will be called by the Payments team to update CCD cases via the Probate Orchestrator service when payments become out of sync.

```
[ ] Yes
[x] No
```


charts/probate-frontend/values.preview.template.yaml
```diff
@@ -45,7 +49,7 @@ rpe-feature-toggle-api:
       primaryIngressHost: ${SERVICE_FQDN}
       prefix: ftr-tgl
     environment:
-      FEATURES_DB_HOST: ${SERVICE_NAME}-ft-db
+      FEATURES_DB_HOST: ${SERVICE_NAME}-postgresql
```
adusumillipraveen: Just get rid of this so that chart picks default one.

PR #794, opened 2019-05-29
PRO-4800 (ex PRI-76): Any Deceased Children

LucaDelBuonoHMCTS:
https://tools.hmcts.net/jira/browse/PRO-4800

### Change description ###
Add "Did any of the children die before the deceased?" question to Intestacy journey

```
[ ] Yes
[X] No
```


app/steps/ui/deceased/anydeceasedchildren/index.js
```diff
@@ -1,12 +1,31 @@
...
 
     static getUrl() {
         return '/any-deceased-children';
     }
+
+    getContextData(req) {
+        const ctx = super.getContextData(req);
+        const formdata = req.session.form;
+        ctx.deceasedName = FormatName.format(formdata.deceased);
+        ctx.deceasedDoD = get(formdata, 'deceased.dod_formattedDate');
```
Wivenhoe-HMCTS: Can you clear deceasedName from the ctx by using the action(), also i would have thought that a field would have been created for deceased.dod_formattedData by default, do we need to create it as deceasedDod as well, if we do, it should be removed in the action() as well.

LucaDelBuonoHMCTS: Added both to action.
I need the DeceasedDoD in the ctx so that then a field is generated to display on the template.

Wivenhoe-HMCTS: Is there not a field called fields.deceased.dod_formattedDate already that you could have used?

PR #890, opened 2019-07-05
PRO-5578: Add applicant name to coversheet

davejones74:
https://tools.hmcts.net/jira/browse/PRO-5578

### Change description ###
Add the applicants name to the coversheet for both intestacy and probate journeys.
The pr change has a related business service change as well.

The frontend change should come either before or during the business service change as the additional application name will be ignored in the business service if the old code is used. However if the business service is updated first the coversheet pdf generation will fail as the application name is a required property. 

```
[ ] Yes
[x] No
```


app/services/IntestacyCoverSheetPdf.js
```diff
@@ -1,13 +1,15 @@
 'use strict';
 
 const IntestacyPdf = require('./IntestacyPdf');
+const {trim, join} = require('lodash');
 
 class IntestacyCoverSheetPdf extends IntestacyPdf {
     post(formdata) {
```
Wivenhoe-HMCTS: Can you use the FormatName.js for the name. thanks.

davejones74: Done for both intestacy and probate journeys.

PR #1113, opened 2020-02-05
Pro 6988 update legal dec flag on tasklist

meganaslattery:

https://tools.hmcts.net/jira/browse/PRO-6988

### Change description ###

FE Bug: executor legal declaration flag not updating correctly

```
[ ] Yes
[x] No
```

### DIFF
--- a/app/middleware/multipleApplications.js
+++ b/app/middleware/multipleApplications.js
@@ -6,7 +6,6 @@ const logger = require('app/components/logger')('Init');
 const ServiceMapper = require('app/utils/ServiceMapper');
 const caseTypes = require('app/utils/CaseTypes');
 const ExecutorsWrapper = require('app/wrappers/Executors');
-const contentWillLeft = require('app/resources/en/translation/screeners/willleft');
 
 const initDashboard = (req, res, next) => {
     const session = req.session;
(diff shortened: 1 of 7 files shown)


app/services/FormData.js
```diff
@@ -60,19 +60,6 @@ class FormData extends Service {
         const fetchOptions = this.fetchOptions({}, 'POST', headers);
         return this.fetchJson(url, fetchOptions);
     }
-
-    getDeclarationStatuses(authToken, serviceAuthorisation, ccdCaseId) {
-        const path = this.replacePlaceholderInPath(this.config.services.orchestrator.paths.declarationStatuses, 'ccdCaseId', ccdCaseId);
```
LucaDelBuonoHMCTS: Good stuff removing this method, shall we also remove the endpoint from config?

app/middleware/multipleApplications.js
```diff
@@ -132,50 +145,39 @@ const getCase = (req, res) => {
...
-
-                        return {
-                            executorName: executor.executorName,
-                            agreed: agreed
-                        };
-                    });
+    result.executors.list
+        .forEach((executor, index) => {
+            if (!executor.isApplicant) {
+                const executorAgreed = get(executor, 'executorAgreed');
```
LucaDelBuonoHMCTS: can you default the get() method to `null` and leave the check below the same instead of using `typeof` and comparing it to `undefined`?

PR #1235, opened 2020-08-18
PRO-7920 enable tests

linusnorton:

https://tools.hmcts.net/jira/browse/PRO-7920

### Change description ###

Enable functional tests in pipeline

```
[ ] Yes
[x] No
```


package.json
```diff
@@ -37,7 +37,7 @@
     "eslint": "NODE_PATH=. eslint .",
     "test:nsp": "NODE_PATH=. nsp check",
     "security": "NODE_PATH=. nsp check",
-    "test:functional": "NODE_ENV=test yarn test-contract && NODE_ENV=test ./bin/run-functional-tests.sh",
+    "test:functional": "NODE_ENV=test yarn test-e2e",
```
velimd, line 40: yarn e2e does this, is there a need for this script?
i think we can just remove it if its just running yarn e2e script and replace whereever this is being used.

velimd: everything else LGTM

linusnorton: it's what the jenkins pipeline calls so it needs to exist

PR #1559, opened 2022-02-22
DTSPB-1493 add PA16 list item to send docs page and coversheet plus r…

itoulou:
https://tools.hmcts.net/jira/browse/DTSPB-1493

### Change description ###
add checklist item to send docs page and coversheet

refactor checklist items generation on send docs template

create mechanism for adding multiple checklist items onto coversheet

```
[ ] Yes
[x] No
```


app/steps/ui/documents/template.html
```diff
@@ -81,7 +16,7 @@
       <h2 class="govuk-heading-m">{{ content.heading1 | safe }}</h2>
       <p class="govuk-body">{{ content.text1 | safe }}</p>
 
-      {{ unorderedList(checklistItems) }}
+      {{ unorderedList(fields.checkListItems.value) }}
```
DelythJustice, line 19: don't we need this as ordered so that it is consistently in same order?

app/wrappers/Applicant.js
```diff
@@ -10,6 +10,15 @@ class Applicant {
...
     }
+
+    applicantIsChild() {
+        return this.formdata.applicant && (this.formdata.applicant.relationshipToDeceased === 'optionChild' || this.formdata.applicant.relationshipToDeceased === 'optionAdoptedChild');
+    }
+
+    spouseIsRenouncing() {
+        return this.formdata.applicant && this.formdata.applicant.spouseNotApplyingReason === 'optionRenouncing';
+    }
+
```
DelythJustice, line 21: I would have named these bit like isSpouseRenouncing, isApplicantChild then obvious you are expecting true or false without going off to look at it

app/steps/ui/documents/template.html
itoulou: I believe the unorderedList widget just sticks whatevers passed in, in an <ul> with bullet points. If we stick it in an <ol> then the list will be numbered which I don't think is what we want. You can alter the  moving around the if statements in the documents page util

app/wrappers/Applicant.js
itoulou: done

DelythJustice (dismissed):

nigeldunne (approved):

nigeldunne (approved):

PR #1567, opened 2022-03-16
DTSPB-2555: fixed hint wording

sanjayparekh-hmcts:
fixed hints wording on screeners

### DIFF
--- a/app/resources/en/translation/screeners/willleft.json
+++ b/app/resources/en/translation/screeners/willleft.json
@@ -3,7 +3,7 @@
   "question": "Did the person who died leave a will?",
   "optionYes": "Yes",
   "optionNo": "No",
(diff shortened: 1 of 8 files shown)


app/resources/en/translation/screeners/willleft.json
```diff
@@ -3,7 +3,7 @@
   "question": "Did the person who died leave a will?",
   "optionYes": "Yes",
   "optionNo": "No",
-
+  "hint": "Wills made according to the laws of England or Wales must be signed in wet ink by:<br><br><ul class=\"govuk-list govuk-list--bullet\"><li>the person who's making the will ('testator')</li><li>2 witnesses</li></ul><br>",
```
nigeldunne, line 6: We're supposed to use &rsquo; instead of '
It's also better to json files for the content and put the html in the template

sanjayparekh-hmcts: ive seen lots of other places that have html in resources, so I just followed that example

sanjayparekh-hmcts: I can change the quotes though....

nigeldunne (approved):

nigeldunne (approved):

PR #1803, opened 2023-02-22
Dtspb 3282 Executor questions update

nilay913:
https://tools.hmcts.net/jira/browse/DTSPB-3282
https://tools.hmcts.net/jira/browse/DTSPB-3377

### Change description ###
Updating Executor pages and error messages in English and Welsh.
Updated the validation regex so names only takes in letters, spaces, apostraphes, hyphens, as per the error message.
Updated the number of executors validation to parseFloat instead of parseInt, so it will return the required error message upon entering a decimal, instead of rounding it to an int (what it used to do). 
Updated relevant test stuff.

```
[ ] Yes
[x] No
```


app/resources/cy/translation/executors/number.json
```diff
@@ -1,16 +1,13 @@
...
+  "question": "Faint o ysgutorion sydd wedi&rsquo;u henwi ar yr ewyllys ac unrhyw codisiliau?",
+  "hintText": "Gwiriwch yr ewyllys a&rsquo;r codisiliau nawr gan fod rhaid ichi gynnwys pob ysgutor a enwir, yn cynnwys y rhai sydd wedi marw.",
+  "executorsNumber": "Nifer yr ysgutorion",
 
   "errors": {
     "executorsNumber": {
-      "required": "Nodwch rif dilys rhwng 1 ac 20",
-      "invalid": "Nodwch rif dilys rhwng 1 ac 20"
+      "required": "Nodwch nifer yr ysgutorion",
+      "invalid": "TO DO - TRANSLATION OF 'Number of executors must be a whole number between 1 and 20'"
```
tomxelliott, line 10: Is this the usual approach rather than adding a TODO comment?

app/steps/ui/executors/number/index.js
```diff
@@ -12,7 +12,7 @@ class ExecutorsNumber extends ValidationStep {
 
     getContextData(req) {
         let ctx = super.getContextData(req);
-        ctx.executorsNumber = ctx.executorsNumber ? parseInt(ctx.executorsNumber) : ctx.executorsNumber;
+        ctx.executorsNumber = ctx.executorsNumber ? parseFloat(ctx.executorsNumber) : ctx.executorsNumber;
```
tomxelliott, line 15: what is the use case for using floating point number parsing here rather than integer parsing?

tomxelliott (dismissed):
Approved with comments

app/resources/cy/translation/executors/number.json
nilay913: the translation should get back today/tomorrow, and will push it then

app/steps/ui/executors/number/index.js
nilay913: Previously it would take the integer in i.e. someone enters 3.912, it would just go through as 3 but not tell the user, so made it parse as float so it would flag it with the correct error message

tomxelliott (approved):

danielbe11 (approved):

PR #1882, opened 2023-08-24
DTSPB-3708  PA: 'Yes' journey ( Died after 1/1/2022)

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/DTSPB-3708


app/journeys/probate.js
IswaryaPepakayala, line 1: As per the scenario 3 ,if they click on Yes on IhtEstatevalued they need to go to SubmittedToHmrc

BWeeks-JusticeGovUK: Fixed both

PR #2364, opened 2025-01-15
DTSPB- 4457 added H2 tag to cookies banner 

abbasiTahir:

https://tools.hmcts.net/jira/browse/DTSPB-4457

```
[ ] Yes
[ x] No
```

### DIFF
--- a/app/views/includes/cookie-banner.html
+++ b/app/views/includes/cookie-banner.html
@@ -5,7 +5,7 @@
         <div class="govuk-cookie-banner__message govuk-width-container cm-cookie-banner__decision">
             <div class="govuk-grid-row">
                 <div class="govuk-grid-column-two-thirds">
-                    <h1 class="govuk-cookie-banner__heading govuk-heading-m">{{ common.cookieBannerTitle }}</h1>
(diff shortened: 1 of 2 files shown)


app/views/includes/cookie-banner.html
```diff
@@ -1,7 +1,7 @@
 <input type="hidden" id="hdnSecurityCookie" value="{{ common.SECURITY_COOKIE }}"/>
 
 <div class="global-cookie-message" id="cm-cookie-banner" hidden>
-    <div class="govuk-cookie-banner" role="region" aria-label="{{ common.cookieBannerTitle }}">
+    <div class="govuk-cookie-banner" role="region" aria-label="{{ common.cookieBannerTitle }}</h2>">
```
tom-saunders-cts, line 4: I think the correct place to be changing this is on line 8? This is simply adding a literal `</h2>` into the `aria-label` attribute which will presumably cause a screen reader to read out something like 'less-than forward-slash h2 greater-than'?

tom-saunders-cts (changes requested):

tom-saunders-cts (dismissed):
I think that this change resolves the issue, but lets get confirmation from someone with some a11y knowledge if we can

tom-saunders-cts (approved):

PR #2380, opened 2025-01-23
DTSPB-4458 Add a11y context for summary page links 

tom-saunders-cts:
See [DTSPB-4458](https://tools.hmcts.net/jira/browse/DTSPB-4458)

### Change description ###
Adds a random string generation method into the nunjucks context, then uses that to provide `id` attributes for templated summary page elements. These `id` values are then used with `aria-labelledby` to add the context for each of the 'Change' or 'Answer' links. This also adds (currently unused) handling for providing specific a11y strings for this context which will use the `aria-label` attribute rather than the auto-generated `aria-labelledby`.

```
[ ] Yes
[X] No
```

### DIFF
--- a/app/resources/cy/translation/dashboard.json
+++ b/app/resources/cy/translation/dashboard.json
@@ -6,6 +6,8 @@
   "tableHeaderCreateDate": "Dyddiad creu",
   "tableHeaderCaseStatus": "Statws",
   "tableHeaderActions": "Camau gweithredu",
+  "unnamedProbateAppl": "WELSH Unnamed probate application",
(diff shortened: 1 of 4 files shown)


tom-saunders-cts:
i do not know enough about our frontend testing to be able to add meaningful tests here. i've asked around the team for advice and haven't had any feedback as yet, but i'm not entirely comfortable about making this sort of change without any test coverage.

MarcusHMCTS (dismissed):

MarcusHMCTS (approved):

PR #2404, opened 2025-02-13
remove request as dependency to replace with Axios

GEvans-HMCTS:
Update dependencies to remove Request and replace it with Axios based on Request Developer recommendations. This should resolve a large amount of CVEs and security issues.

Resolves: https://tools.hmcts.net/jira/browse/DTSPB-3716


test/end-to-end/helpers/TestConfigurator.js
```diff
@@ -2,6 +2,7 @@
 
 const randomstring = require('randomstring');
 const request = require('request');
```
tom-saunders-cts, line 4: are we still transitively including request?

tom-saunders-cts (approved):
this looks reasonable to me

PR #2411, opened 2025-02-21
DTSPB-4565 remove back link from co-applicant pages

GEvans-HMCTS:
Resolves: https://tools.hmcts.net/jira/browse/DTSPB-4565

Removing Back link on multiple pages relating to co-applicant section of journey.


tom-saunders-cts (approved):
i think this should work, but i'm not sure it's really a good pattern to be following longer term - should the steps themselves have a function to determine/report this? Probably a future update though rather than handling here.

tom-saunders-cts (approved):
adding a sixth(?) page

PR #2415, opened 2025-03-05
DTSPB-4586 update helm urls to use oci

GEvans-HMCTS:
resolves https://tools.hmcts.net/jira/browse/DTSPB-4586


tom-saunders-cts (approved):
this seems weird to not update the idam-pr reference to the same repository, but ?

PR #2427, opened 2025-03-25
DTSPB-4449 Add new tab text to IHT links

GEvans-HMCTS:
Resolves: https://tools.hmcts.net/jira/browse/DTSPB-4449


tom-saunders-cts (changes requested):
as mentioned by @FeliTam - likely needs updates in app/resources/cy/translation/screeners/eeestatevalued.json to match

tom-saunders-cts (dismissed):
only question is whether we need to have the `aria-label` here at all (since it's identical to the actual link text)

MarcusHMCTS (approved):

PR #2430, opened 2025-03-31
DTSPB-4459: Initial template change, removing notification banner

jamiealbertelli:

- [ ] update CSS version to force refresh if any changes are introduced

### Change description ###

```
[ ] Yes
[ ] No
```


tom-saunders-cts (commented):
I think the overall changes here are reasonable vs the ask - there are a couple of points where perhaps the formatting of `<span>` elements splitting across lines might be more clear if broken slightly differently.

we've also discussed a potential issue with our a11y handling (the spans mentioned above) with our welsh language support - if a screan reader is using the welsh page it'll read the english prefix then the welsh content. this is probably not the behaviour we should actually be demonstrating, but probably falls outside the scope of this issue.

tom-saunders-cts (approved):

PR #2518, opened 2025-05-16
DTSPB-4727 fix Duplicate names appearing on Legal Declaration

MarcusHMCTS:

[DTSPB-4727](https://tools.hmcts.net/jira/browse/DTSPB-4727)

### Change description ###

check will Name not exist in other Names

```
[ ] Yes
[ ] No
```


app/steps/ui/declaration/index.js
```diff
@@ -107,17 +107,24 @@ class Declaration extends ValidationStep {
     }
 
     collectOtherNames(formdataDeceased) {
```
tom-saunders-cts, line 109: general question - do we check that there are not duplicates between the other entries ?

tom-saunders-cts: there's also some irritating corner cases here where if we have:
| firstNames | lastName |
| --- | --- |
| a | b c |
| a b | c |
the name `a b c` will appear twice in the output - it's unclear whether we can or should do anything about that.

MarcusHMCTS: i don't think don't need check other duplicates

tom-saunders-cts (approved):
based on call this behaviour seems acceptable at this point - we might want to relook at this but it seems low prio

PR #2737, opened 2025-09-03
DTSPB-4942 correct no payment submission

tom-saunders-cts:
See [DTSPB-4942](https://tools.hmcts.net/jira/browse/DTSPB-4942)

### Change description ###
Updates an existing test case to no longer require payment so we don't miss this happening in the future.

Remove the workaround for forced case submission.

Update implementation to correctly cause no-payment cases to submit.

```
[ ] Yes
[x] No
```

This PR picks up the test changes originally from #2711

### DIFF
--- a/app/steps/ui/payment/breakdown/index.js
+++ b/app/steps/ui/payment/breakdown/index.js
@@ -158,8 +158,7 @@ class PaymentBreakdown extends Step {
                     ctx.paymentCreatedDate = paymentResponse.date_created;
                     this.nextStepUrl = () => paymentResponse._links.next_url.href;
                 } else {
-                    logger.info(`forcing case submission for case: ${formdata.ccdCase.id}. ` +
-                        'this will cause a case save failure and redirect the user to the dashboard.');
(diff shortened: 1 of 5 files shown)


FeliTam (dismissed):

charts/probate-frontend/values.preview.template.yaml
```diff
@@ -1,10 +1,10 @@
...
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_NAME}.preview.platform.hmcts.net
-  memoryRequests: 64Mi
-  memoryLimits: 1Gi
-  devmemoryRequests: 64Mi
-  devmemoryLimits: 1Gi
+  memoryRequests: 1Gi
+  memoryLimits: 2Gi
+  devmemoryRequests: 1Gi
+  devmemoryLimits: 2Gi
```
FeliTam, line 7: I believe we won’t need these changes once the memory leak fix goes in, but I’m happy to keep them for now to run the nightly.

FeliTam (approved):

PR #2754, opened 2025-09-11
DTSPB-4947 Fee Increase 2025

MarcusHMCTS:

[DTSPB-4947](https://tools.hmcts.net/jira/browse/DTSPB-4947)

### Change description ###

```
[ ] Yes
[x] No
```


tom-saunders-cts (commented):
similar to the back office - can we convert these to be parameters so we can potentially feature flag these changes? thinking specifically in case we're affected by issues like today where we're unable to deploy due to infra issues.

also is the overseas certified copies changing? i realize it's only used in the test data as we don't seem to have any text telling the user how much they cost but that then raises the question of whether we should?

app/resources/cy/translation/copies/start.json
```diff
@@ -2,6 +2,6 @@
   "title": "Archebu cop&iuml;au ychwanegol o&rsquo;r grant",
   "question": "Archebu cop&iuml;au ychwanegol o&rsquo;r grant",
   "paragraph1": "Mae&rsquo;r ffi gwneud cais yn cynnwys un copi DU o&rsquo;r grant profiant.",
-  "paragraph2": "Mae cop&iuml;au ychwanegol o&rsquo;r grant yn costio &pound;1.50 yr un. Maent yn ddefnyddiol i&rsquo;w hanfon i wahanol sefydliadau ar yr un amser.",
+  "paragraph2": "Mae cop&iuml;au ychwanegol o&rsquo;r grant yn costio &pound;16.00 yr un. Maent yn ddefnyddiol i&rsquo;w hanfon i wahanol sefydliadau ar yr un amser.",
```
tom-saunders-cts, line 5: should this use the same substitution as in `copies/uk`? (ditto in the en version of these strings)

this may be the file/page you were mentioning earlier we don't actually use any more though? if so we should probably remove it rather than leaving dead behaviour around (maybe a different ticket)

app/steps/ui/copies/uk/index.js
```diff
@@ -16,8 +17,8 @@ class CopiesUk extends ValidationStep {
     }
 
     handleGet(ctx, formdata, featureToggles) {
-        ctx.isFeesApiToggleEnabled = featureToggle.isEnabled(featureToggles, 'ft_fees_api');
-
+        ctx.isFtFeesIncrease2025 = featureToggle.isEnabled(featureToggles, 'ft_probate_fee_increase_2025');
```
tom-saunders-cts, line 20: do we need this in the `ctx` ? it's not being used by the template any more?

test/unit/copies/testCopiesUk.js
```diff
@@ -45,14 +45,14 @@ describe('CopiesUk', () => {
...
+        it('should return true when the ft_probate-fee-increase_2025 toggle is set', (done) => {
             const ctxToTest = {};
             const formdata = {};
             const featureToggles = {
-                ft_fees_api: true
+                ft_probate_fee_increase_2025: true
             };
             const [ctx] = CopiesUk.handleGet(ctxToTest, formdata, featureToggles);
-            expect(ctx.isFeesApiToggleEnabled).to.equal(true);
+            expect(ctx.isFtFeesIncrease2025).to.equal(true);
```
tom-saunders-cts, line 55: perhaps this should check that the `copiesUKFee` is correct based on the feature toggle?

test/unit/payment/testPaymentBreakdown.js
```diff
@@ -93,7 +93,7 @@ describe('PaymentBreakdown', () => {
                     applicationfee: 215,
                     applicationvalue: 6000,
                     ukcopies: 1,
-                    ukcopiesfee: 1.50,
+                    ukcopiesfee: 16.00,
                     overseascopies: 2,
                     overseascopiesfee: 3,
```
tom-saunders-cts, line 98: is the overseas copy fee staying at 1.50? if so why would anyone order uk copies vs overseas copies?

tom-saunders-cts (dismissed):
i guess we should get a tech debt ticket preemptively on the backlog to remove the feature toggle after we've gone live

tom-saunders-cts (dismissed):
comments inline for consideration rather than blocking awaiting changes

tom-saunders-cts (dismissed):
i do wonder if we actually need to update all the tests to match the 'correct' payment values as long as the behaviour they're testing is accurate and independent of the actual value of the payment

tom-saunders-cts (approved):

PR #2803, opened 2025-10-02
DTSPB-4811 frontend remove name related validation

justiceia:
https://tools.hmcts.net/jira/browse/DTSPB-4811

### Change description ###
removed validation and unit tests for name related feilds

```
[ ] Yes
[x] No
```


jamiealbertelli:
Hi @justiceia, can we remove the validation for executor current name as well? 

https://github.com/hmcts/probate-frontend/blob/837fc68e2f1fa20ee9afa407e9c049ad818be519/app/steps/ui/executors/currentname/schema.json

jamiealbertelli:
@justiceia Similar to how you have removed the existing error tests, I think you'll need to do the same for the executor current name -> https://github.com/hmcts/probate-frontend/blob/837fc68e2f1fa20ee9afa407e9c049ad818be519/test/component/executors/testExecutorCurrentName.js#L80

jamiealbertelli (dismissed):

tom-saunders-cts (dismissed):

app/steps/ui/executors/names/schema.json
```diff
@@ -4,9 +4,7 @@
   "type": "object",
   "properties": {
     "executorName": {
-      "type": "string",
-      "minLength": 2,
-      "pattern": "^[\\sa-zA-Z'-]+$"
+      "type": "string"
```
tom-saunders-cts: i think we may need this to still be minlength 2 or ccd will refuse it? might need to check that

test/unit/services/testInviteLink.js
```diff
@@ -6,70 +6,63 @@ const InviteLink = require('app/services/InviteLink');
...
+            const fetchJsonStub = sandbox.stub(AsyncFetch, 'fetchJson');
+            sandbox.stub(AsyncFetch, 'fetchOptions').returns(fetchOptions);
+            const formatUrlStub = sandbox.stub(FormatUrl, 'format').returns('/formattedUrl');
 
             inviteLink.get(inviteId);
 
-            expect(inviteLink.log.calledOnce).to.equal(true);
-            expect(inviteLink.log.calledWith('Get invite link')).to.equal(true);
+            expect(logSpy.calledOnce).to.equal(true);
+            expect(logSpy.calledWith('Get invite link')).to.equal(true);
```
tom-saunders-cts: again, i'm wary of testing log calls, but i realise you're just updating existing behaviour

test/unit/services/testInviteLink.js
```diff
@@ -6,70 +6,63 @@ const InviteLink = require('app/services/InviteLink');
...
+                sandbox.stub(AsyncFetch, 'fetchOptions').returns(fetchOptions);
+                const formatUrlStub = sandbox.stub(FormatUrl, 'format').returns('/formattedUrl');
+                const data = [
+                    {Name: 'Test Name', email: 'john.doe@example.com'},
+                    {FirstName: 'John Doe', LastName: 'Smith'}
+                ];
+                inviteLink.post(data, {execObject: true});
+                expect(data[0].Name).to.equal('Test Name');
+                expect(data[1].FirstName).to.equal('John Doe');
+                expect(data[1].LastName).to.equal('Smith');
```
tom-saunders-cts: does `post` modify the input object? if not then isn't this just checking that something we just set up is still the same as we just set up?

tom-saunders-cts (commented):
i think this now works with the changes to the other side not requiring the urlencode - i'm still not entirely sure on whether the minimum lengths are actually enforced by ccd and whether we should remove them if not (or make them match whatever ccd is enforcing if there is some minimum)

tom-saunders-cts (dismissed):
as previous - my only question is on the minimum length handling here vs what ccd is enforcing and whether we should be updating to match that (though we are certainly not causing a problem with the values they currently are here since these restrictions were already present)

tom-saunders-cts (dismissed):
approve with minimal changes from prior

tom-saunders-cts (dismissed):
might be worth confirming we don't need to do similar elsewhere - i don't think it's likely off the top of my head but maybe worth checking the executor links in the "any other execs?" (deletion), the reading for the checkbox for "which execs died?" and "which execs are applying?"

justiceia:
> might be worth confirming we don't need to do similar elsewhere - i don't think it's likely off the top of my head but maybe worth checking the executor links in the "any other execs?" (deletion), the reading for the checkbox for "which execs died?" and "which execs are applying?"

i think they're ok now there was some issues with waf but should be ok now need to see in demo since its a environmental issue

tom-saunders-cts (dismissed):
approving based on minimal change since last approval

tom-saunders-cts (approved):
approving on minimal change since last approval

PR #2925, opened 2025-11-27
DTSPB-3979 change wait for hmrc page button to save & close

tom-saunders-cts:
See [DTSPB-3979](https://tools.hmcts.net/jira/browse/DTSPB-3979)

### Change description ###
Removes the current save & continue button from the form, overrides the default save & exit text link to use govukButton.

Also fixes the formatting on the page for the heading.

```
[ ] Yes
[x] No
```

### DIFF
--- a/app/journeys/intestacy.js
+++ b/app/journeys/intestacy.js
@@ -107,7 +107,7 @@ const stepList = {
         otherwise: 'ProbateEstateValues'
     },
     UniqueProbateCode: 'ProbateEstateValues',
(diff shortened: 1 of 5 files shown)


FeliTam (dismissed):
The changes look sensible to me.
A minor point on the button text, we need to confirm whether it should be "Save and close" or "Save and close your application".
Also, the default.yaml needs to be reverted before merging.

FeliTam (approved):

PR #3232, opened 2026-04-29
[DTSPB-4963] Route direct intestacy applicants to PA16 renunciation flow

sadaque-khan-hmcts:
### JIRA link (if applicable)
See https://tools.hmcts.net/jira/browse/DTSPB-4963

---

### Change description
Route **direct intestacy** cases where a spouse is renouncing to the correct **PA16 give‑up‑administrator‑rights form**, and ensure this is reflected consistently across the application.

This PR:
- Routes eligible spouse‑renunciation intestacy cases to **PA16**
- Aligns behaviour across **Documents**, **Thank You**, and **Coversheet PDF**
- Applies the logic to both **child** and **adopted child** applicants
- Consolidates renunciation logic into shared utilities
- Updates English and Welsh content to use the correct text and links
- Removes outdated or duplicate renunciation checklist items

This change is **functional** with light refactoring for clarity.

---

### Testing done
- Updated and added unit tests for PA16 eligibility and checklist logic
- Updated component tests for Documents and Thank You pages
- Updated coversheet PDF tests
- Verified English and Welsh content
- All tests pass locally and in CI

---

### Impact
Low risk.  
Changes are limited to **direct intestacy spouse‑renunciation flows** and improve correctness and consistency without affecting other journeys.

---

### Security Vulnerability Assessment

CVE Suppression: Are there any CVEs present in the codebase (either newly introduced or pre-existing) that are being intentionally suppressed or ignored by this commit?
- [ ] Yes
- [x] No

---


app/wrappers/Documents.js
```diff
@@ -25,6 +25,14 @@ class Documents {
...
     }
 
+    spouseRenunciationRequired() {
+        return this.deceasedWrapper.hasMarriedStatus() && this.applicantWrapper.isApplicantChild();
+    }
+
+    spouseRenunciationPa16FormRequired() {
+        return this.spouseRenunciationRequired() && this.formdata.caseType === caseTypes.INTESTACY && this.applicantWrapper.isSpouseRenouncing();
+    }
+
```
tom-saunders-cts: this is maybe a little nitpicky (since i think the overall logic might be fine?) but we would only know we specifically require the spouse's renunciation for intestacy 

it's possible that the spouse is listed as an executor but chooses to renounce in a probate application but i don't believe we have any specific handling for that (since it's also possible that the deceased does not name their spouse as one of their executors)

app/utils/DocumentPageUtil.js
```diff
@@ -47,29 +45,34 @@ class DocumentPageUtil {
...
-        if (ctx.spouseRenouncing) {
-            checkListItems.push(content['checklist-item6-spouse-renouncing'].replace('{renunciationFormLink}', config.links.renunciationForm));
-        }
-        if (ctx.isSpouseGivingUpAdminRights) {
-            checkListItems.push(content['checklist-item11-spouse-giving-up-admin-rights-PA16'].replace('{spouseGivingUpAdminRightsPA16Link}', config.links.spouseGivingUpAdminRightsPA16Link));
+        if (ctx.showSpouseRenunciationItem) {
+            const renunciationFormLink = ctx.usePa16RenunciationLink ? config.links.spouseGivingUpAdminRightsPA16Link : config.links.renunciationForm;
+            checkListItems.push(
+                content['checklist-item6-spouse-renouncing'].replace('{renunciationFormLink}', renunciationFormLink)
+            );
```
tom-saunders-cts, line 64: given the comment below (re: only knowing explicit spouse during intestacy) i don't think there can be any circumstance where we would want to use anything but the PA16 link here?

app/utils/DocumentPageUtil.js
```diff
@@ -111,11 +114,9 @@ class DocumentPageUtil {
...
             });
         }
-        if (deceasedWrapper.hasMarriedStatus() && applicantWrapper.isApplicantChild()) {
-            checkListItems.push(this.getCheckListItemTextWithLink(content['checklist-item6-spouse-renouncing'], config.links.renunciationForm));
-        }
-        if (deceasedWrapper.hasMarriedStatus() && applicantWrapper.isApplicantChild() && applicantWrapper.isSpouseRenouncing() && !deceasedWrapper.hasAnyOtherChildren()) {
-            checkListItems.push(this.getCheckListItemTextWithLink(content['checklist-item11-spouse-giving-up-admin-rights-PA16'], config.links.spouseGivingUpAdminRightsPA16Link));
+        if (documentsWrapper.spouseRenunciationRequired()) {
+            const renunciationFormLink = documentsWrapper.spouseRenunciationPa16FormRequired() ? config.links.spouseGivingUpAdminRightsPA16Link : config.links.renunciationForm;
+            checkListItems.push(this.getCheckListItemTextWithLink(content['checklist-item6-spouse-renouncing'], renunciationFormLink));
```
tom-saunders-cts: ditto above comment

test/component/testDocumentsPage.js
```diff
@@ -32,10 +32,9 @@ describe('documents', () => {
             }
         };
         contentData = {
-            ccdReferenceNumber: '1234-1235-1236-1237',
+            ccdReferenceNumber: '1234-1235-1236-1237'
```
tom-saunders-cts, line 35: leave the trailing comma - any future addition to this object does not need to amend this line

test/component/testDocumentsPage.js
```diff
@@ -852,11 +829,49 @@ describe('documents', () => {
                             'checklist-item2-codicils',
                             'checklist-item2-no-codicils',
                             'checklist-item3-codicils-written-wishes',
+                            'checklist-item4-foreign-death-cert',
                             'checklist-item4-interim-death-cert',
+                            'checklist-item4-foreign-death-cert-translation',
```
tom-saunders-cts, line 825: ~~were these actually getting included before? just looking at the changes to test config + the changes made it doesn't seem like the behaviours here should have altered that much?~~ ignore that, github is just showing the addition of the new case afterwards in an unhelpful way

test/component/testDocumentsPage.js
```diff
@@ -852,11 +829,49 @@ describe('documents', () => {
...
+                            'checklist-item9-deed-poll',
+                            'checklist-item10-iht207',
+                            'no-docs-required'
+                        ];
+
+                        testWrapper.testContent(done, contentData, contentToExclude);
+                    });
+            });
+
+            it('test correct content loaded on the page when spouse renounces and applicant is adopted child', (done) => {
```
tom-saunders-cts, line 838: this seems like coverage we need for the current behaviour, but the distinction between child and adopted child is expected to be removed when we release the expanded intestacy - may need to ensure that when this gets merged to release and then picked up into the intestacy branch we're not introducing more odd problems.

test/component/testThankYouPage.js
```diff
@@ -242,7 +244,100 @@ describe('thank-you', () => {
...
+                .end(() => {
+                    const contentData = {
+                        findOutNext: config.links.findOutNext,
+                    };
+
+                    testWrapper.testContent(done, contentData, contentToExclude);
+                });
+        });
+
+        it('test content loaded on the page when spouse renounces and applicant is child', (done) => {
```
tom-saunders-cts, line 263: this is not a criticism of anything you've done (you're just following the pattern of what exists) but i cannot understand what any of these tests are actually testing

app/wrappers/Documents.js
tom-saunders-cts: coming back to this - do we even have a value stored for `isApplicantChild()` or `hasMarriedStatus()` for an application for probate rather than intestacy?

also does `hasMarriedStatus()` cover someone who was in a civil partnership?

app/utils/DocumentPageUtil.js
sadaque-khan-hmcts: From the guidance I was given the fallback is PA15 in anything that doesn't cover PA16 scenario.

app/wrappers/Documents.js
sadaque-khan-hmcts: > this is maybe a little nitpicky (since i think the overall logic might be fine?) but we would only know we specifically require the spouse's renunciation for intestacy
> 
> it's possible that the spouse is listed as an executor but chooses to renounce in a probate application but i don't believe we have any specific handling for that (since it's also possible that the deceased does not name their spouse as one of their executors)

Good points. I checked the wrapper patterns across the repo, and the norm is for wrappers to expose small focused boolean/value methods rather than grouped context objects, so I've kept this in that style for consistency. Also, the intention here was to model these as small business-rules predicates that can still be reused independently if we need the same check elsewhere e.g. the pages steps / as well as the checklist/coversheet generation. 

If this area grows later and we end up with more related form/link decisions in the same renunication flow, that would probably be the point where grouping them into a combined context becomes the better abstraction.

sadaque-khan-hmcts: > coming back to this - do we even have a value stored for `isApplicantChild()` or `hasMarriedStatus()` for an application for probate rather than intestacy?
> 
> also does `hasMarriedStatus()` cover someone who was in a civil partnership?

I checked the underlying data models for this and hasMarriedStatus() is currently checking deceased.maritalStatus === 'optionMarried', where optionMarried is defined in the marital status translations as 'Married or in a civil partnership'. So I don't think we're missing civil partner cases in the current behaviour. I agree though that the method name reads narrower than the actual meaning of the stored value, so that's more of naming clarity issue than a behavioural gap.

test/component/testDocumentsPage.js
sadaque-khan-hmcts: The intent is for the tests to reflect current live behaviour and act as a safety net. Even though this distinction is expected to go away with expanded intestacy, having coverage now means that if something changes unexpectedly, the tests will flag a genuine regression, not a test issue.

When the behaviour is intentionally updated, we can then simplify or remove it if needed.

test/component/testDocumentsPage.js
sadaque-khan-hmcts: These are needed to be excluded in the tests because item 4 is only for death-certificate scenarios, and this is not specified in the setup for the tests.

test/component/testThankYouPage.js
sadaque-khan-hmcts: I can add an assertion to explicitly confirm that **PA16 is being triggered**, but in practice the behaviour and outcome are unchanged. Given the way this test file is structured, all scenarios already follow the same established pattern, and this case is consistent with that approach.

Adding more assertions here wouldn’t increase confidence meaningfully and would mostly duplicate what’s already being validated by the existing setup.

test/component/testDocumentsPage.js
tom-saunders-cts: yeah, this was mostly that i don't think i'm quite back at full speed after my illness and just got completely thrown by what seems like a massive change in a single test but is actually a small change with the addition of a very similar new test afterwards

test/component/testDocumentsPage.js
tom-saunders-cts: agreed - just since our definitions of (e.g.) some of the enums used are often just "check if it's this string"  we've seen issues previously where merges have unintentionally restored removed behavioural paths

test/component/testThankYouPage.js
tom-saunders-cts: i agree - just aware that when these tests start to fail they do so in pretty cryptic ways without obvious indication of the root cause and adding more of them is potentially adding an additional burden for ourselves in the future

test/component/testDocumentsPage.js
```diff
@@ -820,26 +787,25 @@ describe('documents', () => {
...
                         testWrapper.testContent(done, contentData, contentToExclude);
                     });
             });
 
-            it('test correct content loaded on the page when spouse is giving up admin rights and applicant is child', (done) => {
+            it('test correct content loaded on the page when spouse renounces and applicant is child', (done) => {
                 sessionData.deceased = {
                     maritalStatus: 'optionMarried',
-                    anyOtherChildren: 'optionNo'
+                    anyOtherChildren: 'optionYes'
```
tom-saunders-cts, line 802: given the original ticket was to ensure that the behaviour for both any other children yes / no is consistent do we still have coverage for both?

tom-saunders-cts (dismissed):
i think the changes here generally get us to the right end result

tom-saunders-cts (approved):

PR #3256, opened 2026-05-07
DTSPB-5412: Switch from merge() to assign() to prevent unexpected persistance of old data

jamiealbertelli:
[DTSPB-5412](https://tools.hmcts.net/jira/browse/DTSPB-5412)

### Change description ###
Investigate the effect that the merge() call is having on sanitising input and potentially persisting stale case data.

- [ ] Yes
- [x] No


FeliTam (dismissed):
The change seems sensible to me.
Need regression tests to make sure it doesn't accidentally break existing behaviour

FeliTam (approved):

PR #3259, opened 2026-05-07
bump gradle version, dependencies

FeliTam:
### Change description ###
bump gradle version, dependencies

- [ ] Yes
- [x] No


test/end-to-end/helpers/IDAMHelper.js
```diff
@@ -1,39 +1,59 @@
...
-                    throw new Error('TestConfigurator.getBefore: Using proxy - Unable to create user.  Response from IDAM was: ' + response.statusCode);
-                } else {
-                    console.log('User created (via proxy)', options.userDetails);
                 }
+
+                if (response.status !== 201) {
+                    throw new Error(
+                        'TestConfigurator.getBefore: Using proxy - Unable to create user. Response from IDAM was: ' +
+                        response.status
+                    );
```
tom-saunders-cts, line 38: appreciate this is just converting but does the response actually provide any information which might help diagnosis of issues we could include?

tom-saunders-cts (approved):

test/end-to-end/helpers/IDAMHelper.js
FeliTam: Common errors are usually caused by unhealthy IDAM pods, so I am not sure the status text would help much with diagnosis
If there is any useful information we want to capture, we can include it in a separate PR

PR #3355, opened 2026-06-16
DTSPB-5624 Upgrade Postcode lookup with axios

MarcusHMCTS:

DTSPB-5624 Upgrade Postcode lookup with axios

### Change description ###

DTSPB-5624 Upgrade Postcode lookup with axios

- [ ] Yes
- [x] No


app/services/PostcodeLookup.js
```diff
@@ -0,0 +1,54 @@
...
+                    };
+                });
+            })
+            .catch(err => {
+                if (err.response?.status === 401) {
+                    logError('Postcode lookup token is invalid', err);
+                } else {
+                    logError('Postcode lookup service error', err);
+                }
+                return [];
```
tom-saunders-cts: i have to wonder whether if we're getting 500 / timeout errors (for example during the recent service outage) we should be handling that distinctly from simply returning no results - reporting to users that no matching postcode is found suggests they have erred rather than there being a system issue

tom-saunders-cts (changes requested):

app/services/PostcodeLookup.js
tom-saunders-cts: (or even actually if we're failing to authenticate)

MarcusHMCTS: fixed

app/services/PostcodeLookup.js
```diff
@@ -41,12 +41,18 @@ class PostcodeLookup {
...
                     logError('Postcode lookup token is invalid', err);
+                    throw new Error('systemError');
+                } else if (status >= 500 || isTimeout) {
+                    logError('Postcode lookup service outage/timeout', err);
+                    throw new Error('systemError');
                 } else {
                     logError('Postcode lookup service error', err);
+                    throw new Error('systemError');
                 }
-                return [];
```
tom-saunders-cts: have we confirmed that requesting an invalid postcode here doesn't return an error, just an empty response?

tom-saunders-cts (dismissed):
i think this is fine - probably worth confirming that looking up an invalid postcode doesn't unintentionally return a system error message rather than no results before we merge though

FeliTam (approved):

PR #3449, opened 2026-07-28
DTSPB-5821 changed content on 'not entitled to apply' page

justiceia:
https://tools.hmcts.net/jira/browse/DTSPB-5821

### Change description ###
change content en and cy for not entitled page

- [ ] Yes
- [x] No


config/default.yaml
```diff
@@ -240,6 +240,7 @@ links:
   paymentsAccessibilityStatement: 'https://www.payments.service.gov.uk/accessibility-statement/'
   probateEstate: 'https://www.gov.uk/probate-estate'
   findLegalAdvice: 'https://www.gov.uk/find-legal-advice'
+  applicationFormPA1A: 'https://www.gov.uk/government/publications/apply-for-probate-by-post-if-there-is-not-a-will'
```
tom-saunders-cts: two points:
- can we put this with the other `applicationFormPA..` values 
- have we really not had this value in the config already given how many things refer to it? (this is not a specific query, i'm just concerned we've lost something during a merge or similar)

